PR #2, opened 2019-06-09
Add ListFiles function

SmilingNavern:


SmilingNavern:
I am not sure about this approach. Glob() and ReadDir() behaves differently in terms of output. So i provide relative path to files according to args.

sources.go
```diff
@@ -17,6 +19,33 @@ func File(name string) *Pipe {
 	return p.WithReader(f)
 }
 
+// ListFiles contains list of files in `path` as string per line
+// names of files are relative to `path`
+func ListFiles(path string) *Pipe {
+	p := NewPipe()
+
+	if strings.Contains(path, "*") {
```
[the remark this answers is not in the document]
SmilingNavern: I am really not familiar with windows programming but i am going to look into this. Maybe i would need some help)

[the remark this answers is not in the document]
bitfield: I think PR #5 handles this correctly, if you want to compare notes with that one.

SmilingNavern:
I am going to fix this one then.

SmilingNavern:
@bitfield i've updated my PR, thanks for such long delay. I've used ContainsAny and Readme parts from #5 and updated them a little. @abtin Check it please as well.

bitfield:
Great! Could you check that you've addressed all the code review comments so far?

SmilingNavern:
@bitfield i think i have addressed them all except one about Echo(), because i am not sure if this necessary and i am not sure about quality of Readme section, because i am bad in english and readme as well

SmilingNavern:
Actually i thought that ListFiles maybe not a great name. It's more like ListDir(). But i didnt change that because i am not sure. The logic of code is more similiar to something that readdir() do.

bitfield:
Yes, wherever you're creating a pipe from a string, `Echo()` is the thing to use. That means if we want to change the implementation from a `strings.Reader` to something else later, we can.

Don't worry about the Readme text, I'll copyedit that stuff and tweak the wording.

I'm not crazy about the ListFiles() name either, but I can't think of anything better at the moment. `Ls()` is too Unix-specific. `Files()` could be confused with what `Concat()` does—read the contents of the files, not just get the names.

This is where an example program really helps; it shows you how the name of the method will read when combined with others. Let's see if we can come up with an example program which gets a list of files and processes the names in some way.

SmilingNavern:
Okay, i am not sure about name really. Let's stick to `ListFiles()` for now.

I will rewrite to Echo() and think about example.

SmilingNavern:
I've fixed PR for using Echo() to create Pipe.

SmilingNavern:
I've added example. But i think we need more function for files to provide more good examples. Maybe something like `RemoveFile()` or `Gzip()`/`Tar()`

SmilingNavern:
Right now it's not really convenient to do something with this list of path except concating it to match something. I am not sure that `EachLine()` is convenient function for this type of task.

abtin:
Actually I think an iterator function (e.g. Iter, ForEach) can be very
useful.

SmilingNavern:
Yeap, i agree. But i am not sure how to implement it with Pipe concept.

SmilingNavern:
@bitfield hello, are there anything you want me to improve in this PR? I would be glad to hear any additional feedback.

bitfield:
I'll review this as soon as I get a chance! It's looking very good!

README.md
```diff
@@ -375,6 +376,20 @@ fmt.Println(output)
 // Output: contents of file
 ```
 
+## ListFiles
+
+`ListFiles()` Creates a pipe with list of files under the given path or glob.
```
bitfield: How about something like this:

`ListFiles()` lists files, like Unix [`ls`](examples/ls/main.go). It creates a pipe containing all files and directories matching the supplied path specification, one per line. This can be the name of a directory (`/path/to/dir`), the name of a file (`/path/to/file`), or a _glob_ (wildcard expression) conforming to the syntax accepted by [filepath.Match](https://golang.org/pkg/path/filepath/#Match) (`/path/to/*`).

```go
p := script.ListFiles("/tmp/*.php")
files, err := p.String()
if err != nil {
	log.Fatal(err)
}
fmt.Println("found suspicious PHP files in /tmp:")
fmt.Println(files)
```

bitfield: Don't forget to add `ls` to the list of examples at the end of the README.

examples/grep-multiple-files/data/1.log
```diff
@@ -0,0 +1 @@
+hello
```
bitfield: I think maybe let's drop this second example for now—it doesn't really tell us much more than we already know from the `ls` and `grep` examples.

examples/ls/main.go
```diff
@@ -0,0 +1,22 @@
...
+package main
+
+import (
+	"os"
+
+	"github.com/bitfield/script"
+)
+
+func main() {
+	script.ListFiles(os.Args[1]).Stdout()
```
bitfield: We need to handle the case where there are no command-line arguments.

bitfield: Also, I think we need to handle the special cases of `.` and `..`, which Unix `ls` does.

sources.go
```diff
@@ -17,6 +19,33 @@ func File(name string) *Pipe {
 	return p.WithReader(f)
 }
 
+// ListFiles contains list of files in `path` as string per line
```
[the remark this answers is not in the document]
bitfield: Don't forget we need a full doc comment here, explaining everything the method does.

sources.go
```diff
@@ -15,6 +17,27 @@ func Args() *Pipe {
...
+		if err != nil {
+			return NewPipe().WithError(err)
+		}
+		return Echo(strings.Join(fileNames, "\n"))
+	}
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return NewPipe().WithError(err)
+	}
+	var fileNames []string
```
bitfield: Let's use `make` to set the required capacity of this slice when we create it. That's more efficient than creating it with the default capacity and allowing `append` to repeatedly resize it.

sources_test.go
```diff
@@ -26,6 +27,42 @@ func TestFile(t *testing.T) {
 	}
 }
 
+func TestListFilesMultipleFiles(t *testing.T) {
+	t.Parallel()
+	dir := "testdata/multiple_files"
+	want := fmt.Sprintf("%s/1.txt\n%s/2.txt\n%s/3.tar.zip", dir, dir, dir)
```
bitfield: This is a little confusing; let's write it out in full.

sources_test.go
```diff
@@ -26,6 +27,42 @@ func TestFile(t *testing.T) {
...
+	if got != want {
+		t.Errorf("In %q: want %q, got %q", dir, want, got)
+	}
+}
+
+func TestListFilesNonexistent(t *testing.T) {
+	t.Parallel()
+	p := ListFiles("nonexistentpath")
+	if p.Error() == nil {
+		t.Errorf("want error status on listing non-existent path, but got nil")
```
bitfield: This should be `t.Error`.

sources_test.go
```diff
@@ -26,6 +27,42 @@ func TestFile(t *testing.T) {
...
+	}
+}
+
+func TestListFilesGlob(t *testing.T) {
+	t.Parallel()
+	dir := "testdata/multiple_files"
+	want := fmt.Sprintf("%s/1.txt\n%s/2.txt", dir, dir)
```
bitfield: Same again, let's write out the expected string in full.

sources_test.go
```diff
@@ -26,6 +27,42 @@ func TestFile(t *testing.T) {
...
+	p := ListFiles("testdata/multi?le_files/*.txt")
+	got, err := p.String()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if want != got {
+		t.Errorf("In %q: want %q, got %q", dir, want, got)
```
bitfield: Both `want` and `got` already contain the `dir` component, so we don't need to repeat it here.

bitfield:
@SmilingNavern how are you getting along? Sorry I seem to have given you a very long list of changes to make here. It might actually be more efficient for me to make them myself on your branch, then we can agree any final steps that need to happen before merging. What do you think?

If you'd prefer to own the whole thing from start to finish, that's fine too—just let me know.

SmilingNavern:
@bitfield i am going to finish this on weekend, so, please give me a time:) i was busy with work before.

sources_test.go
SmilingNavern: Could you please explain what do you mean by full?

sources_test.go
SmilingNavern: Fixed this one

sources_test.go
SmilingNavern: Fixed this one

sources.go
SmilingNavern: Fixed this as well

examples/ls/main.go
SmilingNavern: i've fixed ls example to hide files starting with single dot. But i don't understand how do you want me to handle double dot? What is special with it?

examples/grep-multiple-files/data/1.log
SmilingNavern: I've dropped it

README.md
SmilingNavern: I've copied your exact description to readme. I am really bad at this, so if there are any issues with readme/docs it would be great if you can fix them.

SmilingNavern:
I've fixed most of issues with my code based on your feedback. If there are any more of them feel free to update current branch and pull request, so we won't delay this PR any longer:)

thanks for your good feedback and sorry if my english isn't great:)

bitfield:
Wonderful, thanks! I’ll review this as soon as I can.

bitfield:
I think we're about there! If you approve my final tweaks, I'll go ahead and merge this. Great job!

SmilingNavern:
Yeah, let's merge this

bitfield:
Thanks again for a great contribution! It's much appreciated.

PR #3, opened 2019-06-10
Add Get() source for retriving web resources

prologic:


sources_test.go
```diff
@@ -111,3 +113,22 @@ func TestArgs(t *testing.T) {
 	}
 
 }
+
+func TestGet(t *testing.T) {
+	defer gock.Off()
```
bitfield: Would you mind very much using plain old `httptest` for this instead? I'm rather pleased with the fact that `script` is currently a zero-import library.

sources.go
```diff
@@ -43,3 +44,14 @@ func Args() *Pipe {
 	}
 	return Echo(s.String())
 }
+
+// Get creates a pipe containing the output of retrieving the web resource
```
bitfield: Can you expand on this a bit (and also update the README)? See the documentation for existing methods for the sort of thing I'm looking for, including error handling, for example.

sources.go
```diff
@@ -43,3 +44,14 @@ func Args() *Pipe {
...
+// given by the supplied url.
+func Get(url string) *Pipe {
+	p := NewPipe()
+	res, err := http.Get(url)
```
bitfield: One thing that concerns me about this is the lack of a timeout in the default HTTP client, for example: https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779

The way I've tackled this in other programs is to have the default behaviour be to use the default client, but allow the user to override this by creating their own client and injecting it somehow. What do you think we should do here?

sources.go
```diff
@@ -43,3 +44,14 @@ func Args() *Pipe {
...
+	if err != nil {
+		return p.WithError(err)
+	}
+	return p.WithReader(res.Body)
```
bitfield: An interesting question here is whether it makes sense (or is possible) to expose the HTTP status code to the user somehow. Usually if you're calling something like Get(), you care if the result is 404, for example. Any thoughts?

sources_test.go
```diff
@@ -111,3 +113,22 @@ func TestArgs(t *testing.T) {
...
+
+	gock.New("http://localhost:8000").
```
bitfield: Can we add a few more test cases here?

sources_test.go
bitfield: For example: https://github.com/bitfield/checkly/blob/master/checkly_test.go#L23

prologic:
> Fantastic PR, thank you! It's much appreciated. I'd love to hear what your use case is for this feature—it's helpful when considering how to design the feature, and it may also suggest useful example programs that we can include.

To be brutally honest; I was just curious about your project, read the [README](/README.md) and implemented what you documented as "suggested new features". I have no use-case at present.

Also the above you've almost copied/pasted word-for-word in every PR I've seen so far :) You might want to think about not doing that so much :D It's a little off putting...

sources_test.go
prologic: So you would rather do a full integration test here instead of a test with mocks (*avoiding any 3rd-party dependencies*)?

sources.go
prologic: I don't see how that's possible right now with the current design of `Pipe`. I don't just want to go around adding random methods to it that have no value for other types of sources, filters and sinks.

One sensible approach would be to more closely mirror the behaviour of `curl` and just return all of the output, headers, status line, body, the lot.

sources_test.go
prologic: Such as? This test fully tests the codepath thus far; I see no value in testing anything else.

sources_test.go
bitfield: No, an integration test would actually call out to the network. I'm suggesting using a local HTTP server, as you do here; just not using the `gock` library.

sources.go
bitfield: Yes, that's an option. I think it's likely to be more useful if `Get()` simply returns the document, without headers. For example, if you're downloading an install script, you could `Get()` it, save it to a file, then `Exec()` it. You don't want to have to strip out headers and status info.

On the other hand, if the download fails altogether (404, for example), there's no point carrying on. We need some way to detect this. How about setting the error status on the pipe if the HTTP status indicates a problem?

sources_test.go
bitfield: There are at least a few things I can think of; redirects and 404 errors, for example. Does Get() follow 3xx redirects?

bitfield:
>  I have no use-case at present.

Well, it's an excellent principle of library design (which I'm trying to follow in this library) not to add a feature just because it's cool, or you think it might be useful, but only when you have an actual requirement for it. The point of this (apart from helping to avoid cluttering up the API with unnecessary things) is that a real-world use case gives you information about how best to design the feature. It's not till you try and use the function yourself that you realise what the API should have been!

bitfield:
> Also the above you've almost copied/pasted word-for-word in every PR I've seen so far :) You might want to think about not doing that so much :D It's a little off putting...

Well, given that every PR needs to be driven by a use case, what would be a better way to communicate this information, or ask this question?

bitfield:
Hey @prologic—are you still interested in working on this? If not, that's fine, but would you mind if I picked up your PR and finished off the changes requested here?

prologic:
By all means :) Go right ahead!

James Mills / prologic

E: prologic@shortcircuit.net.au
W: prologic.shortcircuit.net.au

bitfield:
Closing this one for now, as this feature would benefit from a bit more thought and discussion; when the occasion for that comes up, we can refer to the excellent start made by this PR. Thank you @prologic.

PR #4, opened 2019-06-10
Netcat source filter

prologic:


bitfield:
Closing this for now (there are some more active PRs I need to focus on) but keeping on file for future reference!

Issue #7, opened 2019-06-15
Filters sort and uniq


breml:
First of all, this is a really interesting package, thanks for the effort.

In my scripts I often use the shell commands `sort` and `uniq` so I feel these two would be a good additions for this package.

bitfield:
Yes, I'd like that too! Could you provide a short example program of the kind that you'd like to write in `script` that uses `sort` and `uniq` the way you want to use them? That gives us a concrete design proposal to talk about.

breml:
The use case is normally to process some kind of log file where I try to find out the frequency of some thing. So in `bash` this normally looks like this:

```
cat some.log | grep "the log lines of interest" | awk '{ print $5 }' | sort | uniq -c | sort -rn | head
```

The `awk` part is used to select a particular column (see #6).
With `uniq` I normally use the `-c` to count the frequency.
`sort` is used in two ways, first to sort alphabetically and then to sort by frequency (emited by `uniq`) in reverse order.
In most cases I am only interested in the top X, therefore I use `head`.

bitfield:
Great! Can you turn this into a rough example of the Go program you'd like to write using `script`?

breml:
One challenge is, that the unix tools like `grep`, `tr`, `sort`, `uniq`, etc. have a lot of options and flags, that control the exact behaviour. In order to create an API, that is easy to use and flexible to extend, maybe [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) could be something worth to consider (but this is maybe a topic for another issue).

I think, something like this would be great (already combined with the functional options, which are placed in distinct packages for better readability):

```go
script.Stdin().Match("the log lines of interest").Col(" ", 5).Sort().Uniq(uniq.WithCount()).Sort(sort.Reverse(), sort.Numeric()).Stdout()
```

bitfield:
I'm completely in sympathy with your ideas, but one of the design principles for `script` is that the API should be very simple. The most common things should need zero config. Something like this would be more in keeping with the spirit of the package:

```go
script.Stdin().Match("the log lines of interest").Column(5).Uniq().Sort().CountLines().Stdout()
```

What do you think?

breml:
@bitfield In my opinion, the sample you have put together does not the same as my command, but there is also one simplification, that I like.

### The simplification

[...]

PR #9, opened 2019-06-29
Add First filter

bitfield:
See #7 for a discussion and use case.

### DIFF
--- a/filters.go
+++ b/filters.go
@@ -134,3 +134,26 @@ func (p *Pipe) Concat() *Pipe {
 	})
 	return p.WithReader(io.MultiReader(readers...))
 }
+
+// First reads from the pipe, and returns a new pipe containing only the first N
+// lines. If there is an error reading the pipe, the pipe's error status is also
+// set.
+func (p *Pipe) First(lines int) *Pipe {
+	if p == nil || p.Error() != nil {
+		return p
+	}
+	scanner := bufio.NewScanner(p.Reader)
+	output := strings.Builder{}
+	for i := 0; i < lines; i++ {
+		if !scanner.Scan() {
+			break
+		}
+		output.WriteString(scanner.Text())
+		output.WriteRune('\n')
+	}
+	err := scanner.Err()
+	if err != nil {
+		p.SetError(err)
+	}
+	return Echo(output.String())
+}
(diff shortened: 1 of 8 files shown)


filters.go
```diff
@@ -134,3 +134,26 @@ func (p *Pipe) Concat() *Pipe {
...
+		if !scanner.Scan() {
+			break
+		}
+		output.WriteString(scanner.Text())
+		output.WriteRune('\n')
+	}
+	err := scanner.Err()
+	if err != nil {
+		p.SetError(err)
+	}
```
breml: How does this implementation behave, if you pass a very big file and you use the `First()` filter? With the unix commands, the pipe is closed by `head` after 10 lines, which allows all the commands in front of `head` in the pipe to end as well, without the need of reading the whole file.
Would this work here as well, if we would close the pipe?

filters.go
```diff
@@ -134,3 +134,26 @@ func (p *Pipe) Concat() *Pipe {
 	})
 	return p.WithReader(io.MultiReader(readers...))
 }
+
+// First reads from the pipe, and returns a new pipe containing only the first N
+// lines. If there is an error reading the pipe, the pipe's error status is also
+// set.
+func (p *Pipe) First(lines int) *Pipe {
```
breml: Currently negative values are treated the same as 0. `head` has a special interpretation of negative numbers, which could be added here as well.

filters.go
bitfield: That is a very good point. We don't read any more of the pipe than we need to, but we also don't close it. I'd better add that...

bitfield: Done in 8921a7d

filters.go
bitfield: > GNU head allows negative arguments for -n option, meaning to print all but the last - argument value counted - lines of each input file

That sounds like effectively it would behave like `Last()`—well, we don't have that yet, but when we do, we won't need to worry about this :)

breml: No, it is not like `Last()`, because it prints everything BUT the last n lines, where `Last()` prints ONLY the last n lines. So in fact it is more like an inverted `Last()`.
A similar option is available in `tail` als well.

bitfield: Oh yes, I wasn't reading that properly. Well, it's a neat feature, but I don't have a use case for it—so I'll wait until there is one.

PR #10, opened 2019-07-01
Add Freq() filter

bitfield:
Implements part of what's required for #7.

### DIFF
--- a/filters.go
+++ b/filters.go
@@ -3,10 +3,12 @@ package script
 import (
 	"bufio"
 	"bytes"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"regexp"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


filters.go
```diff
@@ -157,3 +159,38 @@ func (p *Pipe) First(lines int) *Pipe {
...
+// status is also set.
+func (p *Pipe) Freq() *Pipe {
+	if p == nil || p.Error() != nil {
+		return p
+	}
+	var freq = map[string]int{}
+	p.EachLine(func(line string, out *strings.Builder) {
+		freq[line]++
+	})
+	type frequency struct {
```
breml: We could check the len of the map and short cirquit the rest, if we have 0 elements in the map.

filters.go
```diff
@@ -157,3 +159,38 @@ func (p *Pipe) First(lines int) *Pipe {
...
+		freqs = append(freqs, frequency{line, count})
+	}
+	sort.Slice(freqs, func(i, j int) bool {
+		if freqs[i].count == freqs[j].count {
+			return freqs[i].line < freqs[j].line
+		}
+		return freqs[i].count > freqs[j].count
+	})
+	var output strings.Builder
+	for _, item := range freqs {
```
breml: As the slice contains the line with the highest fequency as the first element, we can use the length of the frequency in digits to prefix all lines accordingly with something along these lines (given, we check the len already on line 176):

```
digits := len(strconv.Itoa(freqs[0]))
for _, item := range freqs {
        output.WriteString(fmt.Sprintf(fmt.Sprintf("%% %dd %%s", digits), item.count, item.line))
```
```

File: testdata/freq.input.txt [hunk 3]
```diff
@@ -0,0 +1,11 @@
+apple
```
breml: The test case should at least produce double digit frequencies to test the correct prefixing of the single digit frequencies.

filters.go
bitfield: Yes, that's a good suggestion, but there's not actually a lot to short-circuit. The `for` loop will execute zero times, the `sort.Slice` will be a no-op, and then there'll be no output!

filters.go
bitfield: That's roughly what I ended up doing on the other branch:

```go
	fieldWidth := len(strconv.Itoa(maxCount))
	...
		output.WriteString(fmt.Sprintf("%*d %s", fieldWidth, item.count, item.line))
```

(I was quite pleased to discover the `%*` construct—exactly what I needed!)

testdata/freq.input.txt
bitfield: Yes indeed. Already done on the other branch 👍

filters.go
breml: Great catch, i did not know about `%*`.

bitfield: Neither did I—it's _almost_ undocumented, but it is there in the small print of `fmt`, if you use a large magnifying glass.

PR #15, opened 2019-07-10
Add processes source

SmilingNavern:
Initial try for Processes() source. I am not sure if this is good approach, but it's easiest one. I don't know how to make it protable to other platforms, it's Linux specific right now.

If it's good idea i can do something with it, if it's bad approach then let's close this PR.

### DIFF
--- a/examples/pgrep/main.go
+++ b/examples/pgrep/main.go
@@ -1,0 +1,11 @@
+/* Get pids for processes with name "ssh"
+ */
+package main
+
+import (
+	"github.com/bitfield/script"
+)
+
+func main() {
+	script.Processes().Match("ssh").Stdout()
+}
--- a/sources.go
+++ b/sources.go
@@ -1,8 +1,16 @@
 package script
 
 import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
 	"os"
+	"strconv"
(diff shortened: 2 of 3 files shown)


bitfield:
Thanks for the PR! Yes, I think if we do this, we need to make it as portable as possible. That may mean it's not practical to do—but you might research a little prior art and see what other Go libraries that get process information do.

SmilingNavern:
We can start with Linux-only function for example and make it compile only for linux. For other platforms we can add some placeholder with error. Then if someone wants to add unix/macos/win implementations it can be done later. Personally i wouldn't like to implement other platforms due to my lack of knowledge of those platforms.

My other concerns with this implementation is about errors: you can get error on one process because of permissions but you will still get some list.

bitfield:
It's possible `https://github.com/mitchellh/go-ps` might save us some work—but I think we need to start with an issue and a use case, to help figure out if the feature is needed at all, and if so, how it should be designed. See the [contributor's guide](https://github.com/bitfield/script/blob/master/CONTRIBUTING.md) for more about this.

SmilingNavern:
Yeah, i have read contributor's guide. It's just feature from your list of ideas. Ok i will start issue about this then.

Issue #18, opened 2019-07-12
[Feature Request] Add Last filter


udithprabhu:
Add Last filter which is analogous to tail command

bitfield:
Thank you! Well, you know the drill: please write an example program which uses the feature, and we'll get to work!

udithprabhu:
Last can be used to get the last 10 errors as below

`script.Stdin().Match("error").Last(10).Stdout()`

bitfield:
That's perfectly fine, but not quite what I had in mind. _Why_ would you want this? If you're working on the command line, reading log files, you presumably have `tail` or `less` available. But why would a _program_ want the last X lines of something?

udithprabhu:
One of the usecases that I had was to perform an analysis on recent logs rather than all the logs present in a log file. As you said we can use tail and pipe the result to the go program , but I feel that having this behaviour in the package itself will offer more flexibility.

One more usecase can be that Last paired with Freq can be used to get the least frequent lines.

bitfield:
Analysing 'recent logs' probably needs something a bit more sophisticated than just the last X lines, doesn't it? We wouldn't want to read the same log lines twice, for example.

> One more usecase can be that Last paired with Freq can be used to get the least frequent lines.

Of what, and why? I'm not being purposefully obtuse; I'm trying to dig into what the _use case_ is, which isn't the same as 'an example of using it'. A use case is some real task you need to do.

udithprabhu:
One of the use cases that I encountered recently in my company requires such analysis. The use case was to log the least used control flow path in a critical function, I had thought of using Freq to find it , looking at its implementation , I found that it is sorted in descending order, so with Last the entire implementation can be contained in one go program , otherwise I would have to use tail the output from the go program.

This is what urged me to create the Issue.

One more use case is that finding the Nth line of input such as 

`
p.First(n).Last(1).String()
`

This is present as tail and head in a shell script that I wanted to port to go using scripts

But as the current implementation did not have Last ( at least not a straightforward one ) , I could not port it successfully.

I cannot go more indepth of its needs as the use cases that I have presented are too specific to my organisation, but I assume that such use cases are present for majority of the organisations.

bitfield:
Oh, that's very interesting! Yes, I hadn't thought of using it with `Freq()` to get the _least_ frequent lines. Good idea.

PR #19, opened 2019-07-12
Added Last filter to mimic tail command of shell

udithprabhu:
Added the necessary test function
Added example go program

### DIFF
--- a/README.md
+++ b/README.md
@@ -638,7 +638,6 @@ p := script.File("test.txt").ReplaceRegexp(regexp.MustCompile("Gol[a-z]{1}ng"), 
 script.Stdin().Last(10).Stdout()
 ```
 
-
 # Sinks
 
 Sinks are operations which return some data from a pipe, ending the pipeline.
--- a/README.md
+++ b/README.md
@@ -91,13 +91,13 @@ script.Args().Concat().Match("Error").First(10).AppendFile("/var/log/errors.txt"
 	- [First](#first)
 	- [Freq](#freq)
 	- [Join](#join)
+	- [Last](#last)
 	- [Match](#match)
 	- [MatchRegexp](#matchregexp)
 	- [Reject](#reject)
 	- [RejectRegexp](#rejectregexp)
 	- [Replace](#replace)
 	- [ReplaceRegexp](#replaceregexp)
-	- [Last](#last)
 - [Sinks](#sinks)
 	- [AppendFile](#appendfile)
 	- [Bytes](#bytes)
@@ -582,6 +582,14 @@ fmt.Println(output)
 // Output: hello world\n
 ```
 
+## Last
+
+`Last()` reads its input and passes on the last N lines of it (like Unix [`tail`](examples/tail/main.go)):
+
+```go
+script.Stdin().Last(10).Stdout()
+```
+
 ## Match
 
 `Match()` returns a pipe containing only the input lines which match the supplied string:
@@ -628,14 +636,6 @@ p := script.File("test.txt").Replace("old", "new")
 
 ```go
 p := script.File("test.txt").ReplaceRegexp(regexp.MustCompile("Gol[a-z]{1}ng"), "Go")
-```
-
-## Last
-
-`Last()` reads its input and passes on the last N lines of it (like Unix [`tail`](examples/tail/main.go)):
-
-```go
-script.Stdin().Last(10).Stdout()
 ```
 
 # Sinks
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ script.Args().Concat().Match("Error").First(10).AppendFile("/var/log/errors.txt"
 	- [RejectRegexp](#rejectregexp)
 	- [Replace](#replace)
 	- [ReplaceRegexp](#replaceregexp)
+	- [Last](#last)
 - [Sinks](#sinks)
 	- [AppendFile](#appendfile)
 	- [Bytes](#bytes)
@@ -283,6 +284,7 @@ If you're already familiar with shell scripting and the Unix toolset, here is a 
 | `head`             | [`First()`](#first)                                           |
 | `ls`               | [`ListFiles()`](#listfiles)                                   |
 | `sed`              | [`Replace()`](#replace) / [`ReplaceRegexp()`](#replaceregexp) |
+| `tail`             | [`Last()`](#last)                                             |
 | `uniq -c`          | [`Freq()`](#freq)                                             |
 | `wc -l`            | [`CountLines()`](#countlines)                                 |
 
@@ -628,6 +630,15 @@ p := script.File("test.txt").Replace("old", "new")
 p := script.File("test.txt").ReplaceRegexp(regexp.MustCompile("Gol[a-z]{1}ng"), "Go")
 ```
 
+## Last
+
+`Last()` reads its input and passes on the last N lines of it (like Unix [`tail`](examples/tail/main.go)):
+
+```go
+script.Stdin().Last(10).Stdout()
+```
+
+
 # Sinks
 
 Sinks are operations which return some data from a pipe, ending the pipeline.
@@ -733,6 +744,8 @@ Since `script` is designed to help you write system administration programs, a f
 * [grep](examples/grep/main.go)
 * [head](examples/head/main.go)
 * [echo](examples/echo/main.go)
+* [tail](examples/tail/main.go)
+* [least_freq](examples/least_freq/main.go)
 * [visitors](examples/visitors/main.go)
 * [ls](examples/ls/main.go)
 
--- a/examples/tail/main.go
+++ b/examples/tail/main.go
@@ -1,0 +1,17 @@
+package main
+
+import (
+	"log"
+	"os"
+	"strconv"
+
+	"github.com/bitfield/script"
+)
+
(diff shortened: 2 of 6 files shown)


bitfield:
Thanks for the PR! Don't forget to update the README, and add the new method to the stress-test list, as described in the [contributor's guide](https://github.com/bitfield/script/blob/master/CONTRIBUTING.md#before-submitting-your-pull-request) checklist.

examples/tail/main.go
```diff
@@ -0,0 +1,17 @@
+package main
```
bitfield: Let's make this example program a bit more interesting, and use it with Freq(), as you suggested in the issue discussion. Use it to find the least frequent lines in... something.

filters.go
```diff
@@ -109,6 +109,46 @@ func (p *Pipe) First(lines int) *Pipe {
 	return Echo(output.String())
 }
 
+// Last reads from the pipe, and returns a new pipe containing only the last N
```
bitfield: Note that the methods in this file are in alphabetical order, per the [contributor's guide](https://github.com/bitfield/script/blob/master/CONTRIBUTING.md#style-and-formatting). It just makes it easier to find stuff.

filters.go
```diff
@@ -109,6 +109,46 @@ func (p *Pipe) First(lines int) *Pipe {
...
+// lines. If there is an error reading the pipe, the pipe's error status is also
+// set.
+func (p *Pipe) Last(lines int) *Pipe {
+	if p == nil || p.Error() != nil {
+		return p
+	}
+	scanner := bufio.NewScanner(p.Reader)
+	var stringArr []string
```
bitfield: In Go we don't tend to name our variables after their type (as someone said, you don't name your pets "Cat" and "Dog"). Sure, this is a slice of strings (not an array), but what's it for? The name should clarify the _contents_; the compiler can already tell us the type.

filters.go
```diff
@@ -109,6 +109,46 @@ func (p *Pipe) First(lines int) *Pipe {
...
+	for scanner.Scan(){
+		stringArr = append(stringArr, scanner.Text())
```
bitfield: The problem with this is that we read the _whole_ pipe into memory, before throwing away most of it. Would it be more efficient to use something like a ring buffer that just stores the last X lines?

filters.go
```diff
@@ -109,6 +109,46 @@ func (p *Pipe) First(lines int) *Pipe {
...
+	}
+
+	if len(stringArr) == 0 {
+		return Echo("")
```
bitfield: If you want an empty pipe, it's simpler to use `NewPipe()`.

filters_test.go
```diff
@@ -311,6 +311,49 @@ func TestFirst(t *testing.T) {
 	}
 }
 
+func TestLast(t *testing.T) {
+	t.Parallel()
+	want, err := ioutil.ReadFile("testdata/last10.golden.txt")
+	if err != nil {
+		t.Fatal(err)
+	}
+	input := File("testdata/last.input.txt")
```
bitfield: Since this file is identical to `first.input.txt` I suggest we just re-use that file instead.

filters_test.go
```diff
@@ -311,6 +311,49 @@ func TestFirst(t *testing.T) {
...
+	}
+	if !bytes.Equal(got, want) {
+		t.Errorf("want %q, got %q", want, got)
+	}
+	_, err = ioutil.ReadAll(input.Reader)
+	if err == nil {
+		t.Error("input not closed after reading")
+	}
+	input = File("testdata/last.input.txt")
+	gotZero, err := input.First(0).CountLines()
```
bitfield: I think this should be `Last(0)`, not `First(0)`.

filters_test.go
```diff
@@ -311,6 +311,49 @@ func TestFirst(t *testing.T) {
...
+	}
+	_, err = ioutil.ReadAll(input.Reader)
+	if err == nil {
+		t.Error("input not closed after reading")
+	}
+	want, err = File("testdata/last.input.txt").Bytes()
+	if err != nil {
+		t.Fatal(err)
+	}
+	got, err = File("testdata/last.input.txt").First(100).Bytes()
```
bitfield: Ditto.

udithprabhu:
Thanks for the review,  will send the updated pr in a day or two

examples/tail/main.go
udithprabhu: Will change the example as needed

udithprabhu: Should I create a new example which shows the working of Last along with Freq , because I think that he current example shows the simplest and the most common working of Last , as it emulates the working of tail command.

filters.go
udithprabhu: I can't think of a more efficient way than to read the whole pipe and the current implementation exposes a Reader, and the Reader interface does not have functions to find the length of the content or going to an arbitrary token index.
That is why to find the last n tokens, we need the number of tokens and for that, we need to read the whole Pipe.

Please let me know if I am wrong or if there is a more efficient method to read the content.

examples/tail/main.go
bitfield: What I've tended to do so far is put the minimal example in the README, but in the accompanying example programs, do some slightly more interesting things. After all, the chief value of `script` is the ability to use _combinations_ of methods; if we can produce some working examples which are as close to real sysadmin programs as possible, that's a good thing to aim for.

filters.go
bitfield: Well, to take a trivial example, suppose we just want the very last line. We could do something like the following:

1. Read a line from the input
2. If not the end of input, goto 1.
3. Return the line

The question is really how many lines we need to store in memory, given that we want the last X lines. The logical answer is just X, isn't it?

udithprabhu: Yes, will try to incorporate this logic also

filters.go
```diff
@@ -169,6 +169,44 @@ func (p *Pipe) Join() *Pipe {
 	return Echo(output + terminator)
 }
 
+// Last reads from the pipe, and returns a new pipe containing only the last N
+// lines. If there is an error reading the pipe, the pipe's error status is also
+// set.
+func (p *Pipe) Last(lines int) *Pipe {
```
udithprabhu: I have changed the implementation according to your suggestion. Please review

README.md
```diff
@@ -702,6 +713,8 @@ Since `script` is designed to help you write system administration programs, a f
 * [grep](examples/grep/main.go)
 * [head](examples/head/main.go)
 * [echo](examples/echo/main.go)
+* [tail](examples/tail/main.go)
+* [least frequent](examples/least_freq/main.go)
```
bitfield: Let's make the name `least_freq` to match the directory.

examples/tail/main.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"log"
+	"os"
+	"strconv"
+
+	"github.com/bitfield/script"
+)
+
+func main() {
+	lines, err := strconv.Atoi(os.Args[1])
```
bitfield: What happens when there are no command-line arguments?

filters.go
```diff
@@ -169,6 +169,44 @@ func (p *Pipe) Join() *Pipe {
...
+
+	if lines == 0 {
+		return NewPipe()
+	}
+
+	scanner := bufio.NewScanner(p.Reader)
+	input := make([]string, 0, lines)
+	for scanner.Scan() {
+		if len(input) == lines {
+			input = append(input[1:], scanner.Text())
```
bitfield: This seems like it might be a little inefficient; have you done any performance testing on it? I wonder if this might be a good place to use the [standard library](https://golang.org/pkg/container/ring/#pkg-overview) `ring` package?

filters.go
```diff
@@ -169,6 +169,44 @@ func (p *Pipe) Join() *Pipe {
...
 }
 
+// Last reads from the pipe, and returns a new pipe containing only the last N
+// lines. If there is an error reading the pipe, the pipe's error status is also
+// set.
+func (p *Pipe) Last(lines int) *Pipe {
+	if p == nil || p.Error() != nil {
+		return p
+	}
+
```
bitfield: You'll notice that there are no blank lines within any functions or methods in the `script` package. I tend to think Go code is more readable the more of it you can see at once! For consistency of style, would you mind removing the blank lines?

bitfield:
Also, don't forget to mention in the PR description, "Fixes #18" (or something like that). Then GitHub will automatically link it and close the issue when the PR is merged.

bitfield:
@udithprabhu how's it going? I hope I haven't overwhelmed you with too many changes. If you like, I could pick up this PR and make the requested changes myself, then we can both assess the state of it and see whether it's ready to merge.

udithprabhu:
@bitfield , Sorry for the delay, boggled down by other work, will finish this within 2-3 days if that is fine with you?

bitfield:
Not at all! Take all the time you need, your contribution is much appreciated. Just let me know if I can do anything to help.

udithprabhu:
@bitfield, I think I am more or less done with all the requested changes. please let me know if further tweaking is necessary.

bitfield:
I'll take a look! Thanks very much!

bitfield:
Excellent work! Really outstanding. Thanks very much for the contribution.

Issue #32, opened 2019-10-09
can't exec command with spaces


shumin1027:
can't exec command with spaces such as: `bash -c 'docker ps'`

`p := script.Exec("bash -c 'docker ps'")`

shumin1027:
This is how I solve this problem：

```
package shell

import (
	"bytes"
	"github.com/bitfield/script"
	"os/exec"
)

type PipeWrap struct {
	*script.Pipe
}

func (p *PipeWrap) ToPipe(s string) *script.Pipe {
	return p.Pipe
}

func (p *PipeWrap) Exec(s string) *script.Pipe {
	if p == nil || p.Error() != nil {
		return p.Pipe
	}
	q := NewPipeWrap()
	//args := strings.Fields(s)
	cmd := exec.Command("bash", "-c", s)
	cmd.Stdin = p.Reader
	output, err := cmd.CombinedOutput()
	if err != nil {
		q.SetError(err)
	}
	return q.WithReader(bytes.NewReader(output))
}
func NewPipeWrap() *PipeWrap {
	PipeWrap := new(PipeWrap)
	PipeWrap.Pipe = script.NewPipe()
	return PipeWrap
}

func Exec(s string) *script.Pipe {
	return NewPipeWrap().Exec(s)
}
```

bitfield:
Thanks for the report @shumin1027! PR welcome, from anyone who wants to tackle this.

posener:
The right solution for this problem is to change the API of `Exec` to a slice of string, exactly as the `exec.Command` API is, and pass it as is to the `exec.Command` call.

bitfield:
That's actually a little bit of a pain, which is one reason why we wrap `exec.Command` to be able to pass it a single string.

posener:
> That's actually a little bit of a pain, which is one reason why we wrap `exec.Command` to be able to pass it a single string.

I'm not sure I fully understand what is the pain point. Having the input as a single string?

bitfield:
Yes, `exec.Command` takes a string and a slice of strings as the command line. In shell scripts we can simply write this as one string, so `script` replicates this.

posener:
Right, but then you can't have a single argument which is composed of
multiple words...

bitfield:
Unless you quote it, of course, which is what this issue is really about.

posener:
[...]

PR #33, opened 2019-10-18
Execute a single command where single quotes are needed.

jrswab:
This is in response to issue #32 where the user wanted to enter a command where they specified the shell.

eg. `bash -c 'echo testing'`.

This fix will not handle a command with trailing `&&` at this time but after looking at the readme it seems that may not be the proper use case anyway.

eg. `bash -c 'echo testing' && echo again`

### DIFF
--- a/filters_test.go
+++ b/filters_test.go
@@ -189,6 +189,27 @@ func TestExecFilter(t *testing.T) {
 	if out != "" {
 		t.Error("want exec on erroneous pipe to be a no-op, but it wasn't")
 	}
+
+	// Testing cases with double quotes wrapped in single quoutes.
+	testPipe := NewPipe()
+	testQuotes := []struct {
+		testStrings string
+		want        string
+	}{
+		{`echo 'test'`, "test"},
+		{`echo "a 'b c' d"`, `a 'b c' d`},
+		{`bash -c "echo 'test two'"`, `test two`},
+	}
+	for _, tc := range testQuotes {
+		p = testPipe.Exec(tc.testStrings)
+		got, err = p.String()
+		if err != nil {
+			t.Error(err)
+		}
+		if got != tc.want {
+			t.Errorf("Testing single quotes:\nwant %q, got %q", tc.want, got)
+		}
+	}
 }
 
 func TestJoin(t *testing.T) {
--- a/filters.go
+++ b/filters.go
@@ -115,6 +115,15 @@ func (p *Pipe) Exec(s string) *Pipe {
 	q := NewPipe()
 	args := strings.Fields(s)
 	cmd := exec.Command(args[0], args[1:]...)
+
+	// Check if the command contains a single quote.
+	// such as "bash -c 'echo 10'"
+	if strings.Contains(s, "'") {
+		first := strings.Index(s, "'")
+		last := strings.LastIndex(s, "'")
+		cmd = exec.Command(args[0], args[1], s[first:last+1])
+	}
+
 	cmd.Stdin = p.Reader
 	output, err := cmd.CombinedOutput()
 	if err != nil {
--- a/filters.go
+++ b/filters.go
@@ -113,16 +113,17 @@ func (p *Pipe) Exec(s string) *Pipe {
 		return p
 	}
 	q := NewPipe()
+	var first int
 	args := strings.Fields(s)
-	cmd := exec.Command(args[0], args[1:]...)
 
 	// Check if the command contains a single quote.
-	// such as "bash -c 'echo 10'"
 	if strings.Contains(s, "'") {
-		first := strings.Index(s, "'")
-		last := strings.LastIndex(s, "'")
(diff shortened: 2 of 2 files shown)


bitfield:
Thanks @jrswab! Very nice PR.

Does this fix issue #32 completely? It's not completely clear to me from looking at it whether or not it does.

Could we add a test case which demonstrates the problem reported in #32, which fails with the current version of `script`, but which your patch makes pass?

If this doesn't correctly handle commands with `&&`, is that a new issue that needs to be raised, or should we consider it part of issue #32 and fix it in a separate PR?

jrswab:
> Does this fix issue #32 completely?
It does fix the user's problem completely as well as any other command like it.

> Could we add a test case which demonstrates the problem reported in #32, which fails with the current version of script, but which your patch makes pass?
Sure, do you have a desired way of testing? I was in `filters_test.go` before the PR looking it over. Not sure if you would want the test to be apart of that or `sources_test.go`

> If this doesn't correctly handle commands with &&, is that a new issue that needs to be raised, or should we consider it part of issue #32 and fix it in a separate PR?
If you want the users to have the ability of multiple commands after the initial command like the the one in issue #32 then I'd say it's a new PR. Mainly because that user's direct issue is solved with the code update.

If you want I can step through the code as a comment here and explain since it's not very long.

bitfield:
I'm a bit confused that the issue is about spaces and you're referring to quotes in this PR. It doesn't look as though your patch actually deals with spaces, or am I wrong?

The substantive implementation of Exec is as a filter, so yes, I would add a set of test cases to `TestExecFilter`, along the lines of `TestBasename` in the same file, with various combinations of spaces, quotes, and other edge cases.

jrswab:
> I'm a bit confused that the issue is about spaces and you're referring to quotes in this PR. It doesn't look as though your patch actually deals with spaces, or am I wrong?

Yeah, this is a bit confusing. When I hug into the issue I noticed the code already splits the command on the spaces. The reason why his example was not working was not due to the space but the `'`. When `strings` breaks up the command we end up with `bash` `-c` `'echo` `'test` but in order for the shell to execute `'echo test'` needs to be one parameter.

> The substantive implementation of Exec is as a filter, so yes, I would add a set of test cases to TestExecFilter, along the lines of TestBasename in the same file, with various combinations of spaces, quotes, and other edge cases.

Ok I'll do that this weekend when I get some free time.

bitfield:
Useful comment on #32 which clarifies what we need here:

> Quoting does not help: The input string is passed to [`strings.Field`](https://github.com/bitfield/script/blob/master/filters.go#L116), which is passed to the `exec.Command` in the next line.
> 
> Unfortunately, the strings.Fields does not group into a single field quoted words:
> 
> ```go
> fmt.Println(strings.Join(strings.Fields(`a b "c d" e`), ", "))
> // Output: a, b, "c, d", e
> ```
> 
> Notice that in this example, `"c` is one string returned from `strings.Fields` and `d"` is another, while for the `command.Exec` you would want `c d` as a single string.

jrswab:
> `fmt.Println(strings.Join(strings.Fields('a b "c d" e'), ", "))// Output: a, b, "c, d", e`
>
> Notice that in this example, `"c` is one string returned from `strings.Fields` and `d"` is another, while for the `command.Exec` you would want `c d` as a single string.

Let me set up those tests and add this as an example then tweak the code as needed. It seems to be the same (or a similar) use case that I was aiming to fix with the submitted code.

jrswab:
Just an update:
I am rewriting this to handle a few more cases that will be included in the test. After a bit more thinking I want to make sure it handles more weird cases with quotes than just what #32 was referencing.

bitfield:
Sounds good @jrswab! Let me know if you want another pair of eyes on it.

bitfield:
How are you getting along @jrswab? Want any help?

jrswab:
> How are you getting along @jrswab <https://github.com/jrswab>? Want any help?

I got tied up with life recently. If you want we to push up what I have currently I'd be happy to. That way anyone can help out with it.

bitfield:
Certainly! That would be very helpful, thank you.

filters_test.go
```diff
@@ -189,6 +189,27 @@ func TestExecFilter(t *testing.T) {
...
 		t.Error("want exec on erroneous pipe to be a no-op, but it wasn't")
 	}
+
+	// Testing cases with double quotes wrapped in single quoutes.
+	testPipe := NewPipe()
+	testQuotes := []struct {
+		testStrings string
+		want        string
+	}{
+		{`echo 'test'`, "test"},
```
bitfield: Won't this only work if the `echo` command exists and is in the user's path? Ideally, we wouldn't need to depend on any userland tools for `script` to work; that's part of the reason for writing it. If we just want to test that we correctly assemble command line arguments for a given command, we don't need to actually run the command to do that, do we?

bitfield:
Thanks for your work on this, @jrswab! With your help I've fixed this issue in v0.14.2 via #42.

Issue #35, opened 2019-10-29
List files inside a directory and its subdirectory


mainawycliffe:
We should introduce a method that will list all files in a directory and its subdirectories and return a pipe just like the ListFiles Method. Probably the method should be able to filter hidden files using a `showHidden` argument. I am still yet to come up with a decent name for the method, I 
am open to suggestions. 

```go
func FindFiles(path string, showHidden bool) *Pipe {
  // content
}
```

bitfield:
Sounds good, @mainawycliffe! Let's punt the question of hidden files for the moment and try to come up with the simplest thing that could possibly work.

`FindFiles` sounds like a good name. Given the following directory tree:

```
foo/
-- file1.txt
-- file2.txt
-- bar/
----file3.txt
```

I guess the output of FindFiles should be:

```go
foo/file1.txt
foo/file2.txt
foo/bar/file3.txt
```

What do you think?

mainawycliffe:
> Sounds good, @mainawycliffe! Let's punt the question of hidden files for the moment and try to come up with the simplest thing that could possibly work.

Yes, we should keep things as simple as possible. 

> FindFiles sounds like a good name. Given the following directory tree:
> 
> foo/
> -- file1.txt
> -- file2.txt
> -- bar/
> ----file3.txt
> I guess the output of FindFiles should be:
> 
> foo/file1.txt
> foo/file2.txt
> foo/bar/file3.txt
> What do you think?

Yeah, the output should be exactly like that.

bitfield:
Fixed by #36!

Issue #23, opened 2019-07-17
Implement 'find TODOs' example


bitfield:
https://spin.atomicobject.com/2019/07/17/script-find-todos-code/#.XS8eSdFwrY0.hackernews

mainawycliffe:
This is interesting to me, I can take it up and attempt to find a solution.

bitfield:
I'm looking forward to seeing what you come up with @mainawycliffe! If you're not sure how to get started, here's what I do: I write the program that I'd like to write with `script` first, and then worry about how to implement it! That means by the time I come to implement a method, I already know what arguments it takes, what it does, and what it returns.

bitfield:
@mainawycliffe any news on this? Let me know if you'd like some help.

mainawycliffe:
@bitfield not yet, been working on it mainly weekends. I have a working script and I just need to expand it a little bit so it can work with more comments type and able to get complete todo messages when they span across multiple lines. I hope to have a working prototype over the next weekend and then we can see where we can make improvements.

bitfield:
Sounds great @mainawycliffe! Feel free to share your work in progress with me via john@bitfieldconsulting.com, I may be able to make some suggestions.

mainawycliffe:
I will, Thanks.

Regards
Maina Wycliffe

________________________________

bitfield:
I think this is effectively solved in #36, isn't it?

PR #36, opened 2019-10-30
Add Findfiles func to find files recursively

mainawycliffe:
Add a `FindFiles` func that will find files recursively in all directories (#35)

**NOTE:** 

I wanted the method to accept glob, but its behaving strangely in go. When you pass `dir/**/*.txt`, it only finds files inside the subdirectories ignoring all files in the root directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,10 +58,20 @@ What's that? You want to append that output to a file instead of printing it to 
 script.Args().Concat().Match("Error").First(10).AppendFile("/var/log/errors.txt")
 ```
 
+# Want some help with Go—or anything else?
+
+Not content with maintaining this library, John Arundel, of [Bitfield Consulting](https://bitfieldconsulting.com), is a highly experienced Go trainer and mentor who can teach you Go from scratch, take you beyond the basics, or even help you reach complete mastery of the Go programming language. See [Learn Go with Bitfield](https://bitfieldconsulting.com/golang) for details, or email go@bitfieldconsulting.com to find out more.
+
+> John's Golang mentoring has helped me build confidence and fill in gaps in my knowledge. It has provided an immeasurable amount of help and guidance, and as a result I'm applying for my dream job as an SRE!<br />
+—Melina Boutierou
+
+John is also a [Kubernetes and cloud infrastructure consultant](https://bitfieldconsulting.com/kubernetes) and the author of the book [Cloud Native DevOps with Kubernetes](https://amzn.to/2PEPTjc). If John can help you with your infrastructure or DevOps projects, [get in touch](https://bitfieldconsulting.com/contact)! He'd love to hear from you.
+
 # Table of contents<!-- omit in toc -->
 - [What is `script`?](#what-is-script)
 - [How do I import it?](#how-do-i-import-it)
 - [What can I do with it?](#what-can-i-do-with-it)
+- [Want some help with Go—or anything else?](#want-some-help-with-goor-anything-else)
 - [How does it work?](#how-does-it-work)
 - [Everything is a pipe](#everything-is-a-pipe)
 - [What use is a pipe?](#what-use-is-a-pipe)
--- a/README.md
+++ b/README.md
@@ -347,7 +347,7 @@ fmt.Println(output)
 `Exec()` runs a given command and creates a pipe containing its combined output (`stdout` and `stderr`). If there was an error running the command, the pipe's error status will be set.
 
 ```go
-p := script.Exec("echo hello")
+p := script.Exec("bash -c 'echo hello'")
 output, err := p.String()
 fmt.Println(output)
 // Output: hello
@@ -827,6 +827,7 @@ Since `script` is designed to help you write system administration programs, a f
 
 * [cat](examples/cat/main.go) (copies stdin to stdout)
 * [cat 2](examples/cat2/main.go) (takes a list of files on the command line and concatenates their contents to stdout)
+* [execute](examples/execute/main.go)
 * [grep](examples/grep/main.go)
 * [head](examples/head/main.go)
 * [echo](examples/echo/main.go)
--- a/README.md
+++ b/README.md
@@ -554,7 +554,7 @@ p := File("filelist.txt").Concat()
 p := Exec("ls /var/app/config/").Concat().Stdout()
 ```
 
-Each input file will be closed once it has been fully read.
+Each input file will be closed once it has been fully read. If any of the files can't be opened or read, `Concat()` will simply skip these and carry on, without setting the pipe's error status. This mimics the behaviour of Unix `cat`.
 
 ## Dirname
 
--- a/README.md
+++ b/README.md
@@ -422,16 +422,11 @@ script.IfExists("/foo/bar").Exec("/usr/bin/yada")
 
 ## FindFiles
 
-`FindFiles()` finds all files in a directory and its subdirectories, like Unix [`find`](examples/find/main.go). It creates a pipe containing all files inside a given directory and its subdirectories, one per line.
+`FindFiles()` lists all files in a directory and its subdirectories recursively, like Unix [`find -type f`](examples/find/main.go).
 
 ```go
-p := script.FindFiles("/tmp")
-files, err := p.String()
-if err != nil {
-	log.Fatal(err)
-}
-fmt.Println("The /tmp directory contains the following:")
-fmt.Println(files)
+script.FindFiles("/tmp").Stdout()
+// lists all files in /tmp and its subtrees
 ```
 
 ## ListFiles
--- a/README.md
+++ b/README.md
@@ -289,6 +289,7 @@ If you're already familiar with shell scripting and the Unix toolset, here is a 
 | `grep`             | [`Match()`](#match) / [`MatchRegexp()`](#matchregexp)         |
 | `grep -v`          | [`Reject()`](#reject) / [`RejectRegexp()`](#rejectregexp)     |
 | `head`             | [`First()`](#first)                                           |
+| `find`             | [`FindFiles`](#findfiles)                                     |
 | `ls`               | [`ListFiles()`](#listfiles)                                   |
 | `sed`              | [`Replace()`](#replace) / [`ReplaceRegexp()`](#replaceregexp) |
 | `tail`             | [`Last()`](#last)                                             |
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ John is also a [Kubernetes and cloud infrastructure consultant](https://bitfield
 	- [Dirname](#dirname)
 	- [EachLine](#eachline)
 	- [Exec](#exec-1)
+	- [ExecForEach](#execforeach)
 	- [First](#first)
 	- [Freq](#freq)
 	- [Join](#join)
@@ -305,6 +306,7 @@ If you're already familiar with shell scripting and the Unix toolset, here is a 
 | `tail`             | [`Last()`](#last)                                             |
 | `uniq -c`          | [`Freq()`](#freq)                                             |
 | `wc -l`            | [`CountLines()`](#countlines)                                 |
+| `xargs`            | [`ExecForEach()`](#execforeach)                               |
 
 # Sources, filters, and sinks
 
@@ -603,6 +605,19 @@ fmt.Println(output)
 // Output: hello world
 ```
 
+## ExecForEach
+
+ExecForEach runs the supplied command once for each line of input, and returns a pipe containing the output, like Unix `xargs`.
+
+The command string is interpreted as a Go template, so `{{.}}` will be replaced with the input value, for example.
+
+The first command which results in an error will set the pipe's error status accordingly, and no subsequent commands will be run.
+
+```go
+// Execute all PHP files in current directory and print output
+script.ListFiles("*.php").Exec("php {{.}}").Stdout()
+```
+
 ## First
 
 `First()` reads its input and passes on the first N lines of it (like Unix [`head`](examples/head/main.go)):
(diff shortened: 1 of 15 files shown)


bitfield:
Are you ready for a review yet @mainawycliffe or are you still working on it?

mainawycliffe:
I am ready for a review. I think we can add glob to `FindFile` func if the situation changes in the future.

sources.go
```diff
@@ -83,6 +83,31 @@ func ListFiles(path string) *Pipe {
 	return Slice(fileNames)
 }
 
+// FindFiles creates a pipe containing the files inside a directory and all its subdirectories
+// matching the supplied path, one per line.
+func FindFiles(path string) *Pipe {
+
```
bitfield: It's house style in this repo not to use blank lines within functions—or, at least, to do so incredibly sparingly. The trouble with blank lines is, where do you stop? Eventually every other line is blank and the code is twice as long!

sources.go
```diff
@@ -83,6 +83,31 @@ func ListFiles(path string) *Pipe {
...
+	fileNames := make([]string, 0)
```
bitfield:
```suggestion
	var fileNames []string
```

sources.go
```diff
@@ -83,6 +83,31 @@ func ListFiles(path string) *Pipe {
 	return Slice(fileNames)
 }
 
+// FindFiles creates a pipe containing the files inside a directory and all its subdirectories
```
bitfield: Functions are in alphabetical order in this file, just to keep some semblance of organization—so this should actually go before `ListFiles`.

sources_test.go
```diff
@@ -155,6 +155,40 @@ func TestListFilesGlob(t *testing.T) {
 	}
 }
 
+func TestFindFilesMultipleFiles(t *testing.T) {
```
bitfield: The tests are also in alphabetical order!

sources_test.go
```diff
@@ -155,6 +155,40 @@ func TestListFilesGlob(t *testing.T) {
...
+	t.Parallel()
+	dir := "testdata/multiple_files"
+	want := fmt.Sprintf("%s/1.txt\n%s/2.txt\n%s/3.tar.zip\n", dir, dir, dir)
+	got, err := ListFiles(dir).String()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got != want {
```
bitfield: I try to use `cmp.Equal` and `cmp.Diff` wherever possible in the tests, because the output is so much more helpful than simply comparing strings. See other tests for how to use these.

sources_test.go
```diff
@@ -155,6 +155,40 @@ func TestListFilesGlob(t *testing.T) {
 	}
 }
 
+func TestFindFilesMultipleFiles(t *testing.T) {
+	t.Parallel()
+	dir := "testdata/multiple_files"
+	want := fmt.Sprintf("%s/1.txt\n%s/2.txt\n%s/3.tar.zip\n", dir, dir, dir)
+	got, err := ListFiles(dir).String()
```
bitfield: We probably want to call `FindFiles` rather than `ListFiles` here, don't we? Or another way to think it about it, if there are no subdirectories, then the result of `FindFiles` should be the same as that of `ListFiles`. But I prefer having an explicit expectation, as you do here.

sources_test.go
```diff
@@ -155,6 +155,40 @@ func TestListFilesGlob(t *testing.T) {
...
+	}
+	if got != want {
+		t.Errorf("Want %q, got %q", want, got)
+	}
+}
+
+func TestFindFilesMultipleFilesWithSubdirectory(t *testing.T) {
+	t.Parallel()
+	dir := "testdata/multiple_files_with_subdirectory"
+	want := fmt.Sprintf("%s/1.txt\n%s/2.txt\n%s/3.tar.zip\n%s/dir/1.txt\n%s/dir/2.txt\n", dir, dir, dir, dir, dir)
```
bitfield: Can we add a 'hidden' file (one whose name starts with a dot) to these test cases? FindFiles should find it.

sources.go
bitfield: Also, you'll notice that comments are wrapped to about 80 columns (I use the Rewrap extension in VS Code for this, which is very handy).

README.md
```diff
@@ -81,6 +81,7 @@ script.Args().Concat().Match("Error").First(10).AppendFile("/var/log/errors.txt"
 		- [Error output](#error-output)
 	- [File](#file)
 	- [IfExists](#ifexists)
+	- [FindFiles](#findfiles)
```
bitfield: Shall we add this to the list of Unix equivalents, as well, as the equivalent of `find`?

sources_test.go
```diff
@@ -97,6 +97,40 @@ func TestFile(t *testing.T) {
...
+func TestFindFilesMultipleFiles(t *testing.T) {
+	t.Parallel()
+	dir := "testdata/multiple_files"
+	want := fmt.Sprintf("%s/1.txt\n%s/2.txt\n%s/3.tar.zip\n", dir, dir, dir)
+	got, err := FindFiles(dir).String()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !cmp.Equal(want, got) {
+		t.Errorf("Want %q, got %q", want, got)
```
bitfield:
```suggestion
		t.Error(cmp.Diff(want, got))
```

sources_test.go
```diff
@@ -97,6 +97,40 @@ func TestFile(t *testing.T) {
...
+func TestFindFilesMultipleFilesWithSubdirectory(t *testing.T) {
+	t.Parallel()
+	dir := "testdata/multiple_files_with_subdirectory"
+	want := fmt.Sprintf("%s/1.txt\n%s/2.txt\n%s/3.tar.zip\n%s/dir/.hidden\n%s/dir/1.txt\n%s/dir/2.txt\n", dir, dir, dir, dir, dir, dir)
+	got, err := FindFiles(dir).String()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !cmp.Equal(want, got) {
+		t.Errorf("Want %q, got %q", want, got)
```
bitfield:
```suggestion
		t.Error(cmp.Diff(want, got))
```

README.md
mainawycliffe: Yea, I think so, although a bit limited as we don't have glob support for find file.

mainawycliffe:
I am Sorry, I accidentally merged my branch on #23 and became too tangled to undo. I think the [example](https://github.com/mainawycliffe/script/tree/findfiles/examples/todo) is also for a review, so you can take a look at that too.

examples/todo/main.go
```diff
@@ -0,0 +1,139 @@
...
+	listPath := "."
+	if len(os.Args) > 1 {
+		listPath = os.Args[1]
+	}
+	filterFiles := regexp.MustCompile(`^\..*|/\.`)
+	files := script.FindFiles(listPath).RejectRegexp(filterFiles)
+	content := files.EachLine(func(filePath string, builderFile *strings.Builder) {
+		p := script.File(filePath)
+		lineNumber := 1
+		// keep track of comments0
```
bitfield:
```suggestion
		// keep track of comments
```

examples/todo/main.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"github.com/bitfield/script"
+)
+
+func main() {
+	listPath := "."
+	if len(os.Args) > 1 {
+		listPath = os.Args[1]
+	}
+	filterFiles := regexp.MustCompile(`^\..*|/\.`)
```
bitfield: Maybe worth a comment to explain what this regexp is filtering out?

examples/todo/main.go
```diff
@@ -0,0 +1,139 @@
+package main
```
bitfield: General comment: I think this is a really good example program, certainly the best we have for `script`. Actually, it may be a little _too_ good! The aim is to show how to use `script` to do useful things, but it's always difficult to write an example which shows what you want to show _and_ does something useful, without obscuring the main point.

You might like to take this program as is and give it its own repo (it's a useful tool!), and then slightly shorten this example to focus only on the `script` parts. For example, we could change this to just match the string `TODO` in any line, rather than detecting comment blocks, which is hard!

examples/todo/examples.text
```diff
@@ -0,0 +1,94 @@
+package script
```
bitfield: It might make it easier for readers to understand this if the example source code we're using to find todos in is *not*, in fact, a program for finding todos! Any other simple program would do, but this is potentially confusing.

examples/todo/main.go
mainawycliffe: I have simplified the example to only find todo existence in each line of code.

mainawycliffe:
I have created a new repo for the find Todos program [here](https://github.com/mainawycliffe/script-find-todos). This is actually a good idea, as I can work to improve it and improve comment detection on the repo.

examples/todo/main.go
```diff
@@ -0,0 +1,47 @@
...
+		listPath = os.Args[1]
+	}
+	// filter hidden directories and files
+	filterFiles := regexp.MustCompile(`^\..*|/\.`)
+	files := script.FindFiles(listPath).RejectRegexp(filterFiles)
+	content := files.EachLine(func(filePath string, builderFile *strings.Builder) {
+		p := script.File(filePath)
+		lineNumber := 1
+		p.EachLine(func(str string, build *strings.Builder) {
+			findTodo, err := hasTodo([]byte(str))
```
bitfield: Could we use `strings.Contains` here to simplify? This saves calling a function, building a regexp, and translating back and forth from `string` to `[]byte`.

sources.go
```diff
@@ -53,6 +53,25 @@ func File(name string) *Pipe {
...
+	err := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsDir() {
+			fileNames = append(fileNames, path)
+		}
+		return nil
+	})
+	if err != nil {
```
bitfield: By the way, a common pattern is to rewrite something like:

```go
err := foo()
if err != nil {
```

as:

```go
if err := foo(); err != nil {
```

But you may feel that the function call is too long for this style to be readable. If so, I trust your judgment!

examples/todo/main.go
mainawycliffe: Yes, we could use strings.Contains to simplify the example further and for the purpose of the example, it would work.

sources.go
mainawycliffe: For this to be readable, we would have to extract the function to a variable, then we can use the pattern. And I think that could be more readable.

examples/todo/examples.text
bitfield: Will you substitute another program, then? We could use one of the other `script` examples, perhaps.

examples/todo/main.go
```diff
@@ -0,0 +1,32 @@
...
+		listPath = os.Args[1]
+	}
+	// filter hidden directories and files
+	filterFiles := regexp.MustCompile(`^\..*|/\.`)
+	files := script.FindFiles(listPath).RejectRegexp(filterFiles)
+	content := files.EachLine(func(filePath string, builderFile *strings.Builder) {
+		p := script.File(filePath)
+		lineNumber := 1
+		p.EachLine(func(str string, build *strings.Builder) {
+			findTodo := strings.Contains(str, "todo")
```
bitfield:
```suggestion
			if strings.Contains(str, "todo") {
```

examples/todo/main.go
```diff
@@ -0,0 +1,32 @@
...
+			if findTodo {
+				builderFile.WriteString(fmt.Sprintf("%s:%d %s \n", filePath, lineNumber, strings.TrimSpace(str)))
```
bitfield: Rather than calling `fmt.Sprintf` and then `builderFile.WriteString`, we can use `builderFile` as an `io.Writer` directly:

```suggestion
				fmt.Fprintf(builderFile, "%s:%d %s \n", filePath, lineNumber, strings.TrimSpace(str)))
```

examples/todo/main.go
mainawycliffe: Yeah, this makes sense

mainawycliffe:
no problem @bitfield, that happens all the time.

examples/find/main.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/bitfield/script"
+)
+
+func main() {
+	listPath := "."
+	if len(os.Args) > 1 {
+		listPath = os.Args[1]
+	}
+	_, err := script.FindFiles(listPath).Stdout()
```
bitfield: In general it's not much use trapping an error writing to standard out. Even if we get one, what can we do about it? :)

```suggestion
	script.FindFiles(listPath).Stdout()
```

sources_test.go
```diff
@@ -97,6 +97,40 @@ func TestFile(t *testing.T) {
 	}
 }
 
+func TestFindFilesMultipleFiles(t *testing.T) {
+	t.Parallel()
```
bitfield: Instead of three different tests, could you write this as a single table-driven test? Have a look at [TestExec](https://github.com/bitfield/script/blob/master/sources_test.go#L45) for an example. This avoids repeating the same test code, and makes it easier to add new test cases if we need to.

mainawycliffe:
Taken care of the changes requested above.

bitfield:
Great work, @mainawycliffe! Do you mind if I give this a final pass for style, consistency, squash commits, etc?

mainawycliffe:
No, I don't.

bitfield:
Thanks again for the feature, and all your patient collaboration @mainawycliffe.

mainawycliffe:
Thanks too, learned a lot. Any new issue that I can work on?

bitfield:
I think what we could most use is some substantial, non-trivial, useful programs written using `script`. If they're easy to write, they'll make great examples and demos. If they're not, that may be pointing us towards improvements and additions in `script` that would make such programs easier. See what you can come up with!

Issue #39, opened 2020-03-12
calc md5hash of file


hariboyapally:
there is always a use case to calculate md5hash of a  file and compare it with others. Having a single liner will definitely help several use cases.

bitfield:
Hey @hredd00, great idea! Thank you!

Since MD5 is not cryptographically secure, maybe we should think about using a newer and more secure hash algorithm, such as SHA-256?

hredd00:
Agreed that we should have a secure hash like SHA-256; however, in one of my use cases I frequently check file integrity with md5 and achieving the same with SHA-256 is more CPU-intensive.
But Yes; another one-liner for SHA256 :)
Thoughts?

bitfield:
Yes, I think that's probably what most people use it for. My concern is, though, that somebody will see this and use it for some security-sensitive application, which it absolutely isn't suitable for. I wonder just how much more expensive (per megabyte) SHA-256 is. Are there any figures on this?

bitfield:
(For example, think about a tool like `tripwire`, which calculates hash sums for installed binaries and uses these to verify that the binary hasn't been modified. MD5/SHA-1 is absolutely unsuitable for this, since you can construct a hacked binary which hashes to the required value.)

natea123:
Hello,

Very much new to open-source contributing but learning Go and would like to take a crack at this. Is the goal to just have a one-liner that returns the sha256 hash of a selected file? I noticed there is a sha256 package [1], would it be viable to use that package for implementation?

[1] https://golang.org/pkg/crypto/sha256/

thomaspoignant:
The `SHA256Sum` is merged, @hredd00 you can use it to calculate files hashes.

etimoz:
Is the md5 hashsum feature still wanted? Would work on it if yes, otherwise I think this issue can be closed.

Correct my if I'm wrong but I think it could still be used to check file integrity see this discussion on [stackoverflow](https://security.stackexchange.com/questions/87375/why-are-md5-and-sha-1-still-used-for-checksums-and-certificates-if-they-are-call).

bitfield:
[...]

PR #40, opened 2020-03-17
calc SHA-256 hash of file

thomaspoignant:
Add a `CheckSum()` sinks func that will return the checksum of a file. (https://github.com/bitfield/script/issues/39).

The checksum is a **SHA-256** hash of the file.
The result is a `string` to be readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,27 +284,27 @@ func main() {
 
 If you're already familiar with shell scripting and the Unix toolset, here is a rough guide to the equivalent `script` operation for each listed Unix command.
 
-| Unix / shell          | `script` equivalent                                           |
-| --------------------- | ------------------------------------------------------------- |
-| (any program name)    | [`Exec()`](#exec)                                             |
-| `[ -f FILE ]`         | [`IfExists()`](#ifexists)                                     |
-| `>`                   | [`WriteFile()`](#writefile)                                   |
-| `>>`                  | [`AppendFile()`](#appendfile)                                 |
-| `$*`                  | [`Args()`](#args)                                             |
-| `basename`            | [`Basename()`](#basename)                                     |
-| `cat`                 | [`File()`](#file) / [`Concat()`](#concat)                     |
-| `cut`                 | [`Column()`](#column)                                         |
-| `dirname`             | [`Dirname()`](#dirname)                                       |
-| `echo`                | [`Echo()`](#echo)                                             |
-| `grep`                | [`Match()`](#match) / [`MatchRegexp()`](#matchregexp)         |
-| `grep -v`             | [`Reject()`](#reject) / [`RejectRegexp()`](#rejectregexp)     |
-| `head`                | [`First()`](#first)                                           |
-| `ls`                  | [`ListFiles()`](#listfiles)                                   |
-| `sed`                 | [`Replace()`](#replace) / [`ReplaceRegexp()`](#replaceregexp) |
-| `openssl dgst -sha256`| [`SHA256Sum()`](#replace)                    |
-| `tail`                | [`Last()`](#last)                                             |
-| `uniq -c`             | [`Freq()`](#freq)                                             |
-| `wc -l`               | [`CountLines()`](#countlines)                                 |
+| Unix / shell       | `script` equivalent                                           |
+| ------------------ | ------------------------------------------------------------- |
+| (any program name) | [`Exec()`](#exec)                                             |
+| `[ -f FILE ]`      | [`IfExists()`](#ifexists)                                     |
+| `>`                | [`WriteFile()`](#writefile)                                   |
+| `>>`               | [`AppendFile()`](#appendfile)                                 |
+| `$*`               | [`Args()`](#args)                                             |
+| `basename`         | [`Basename()`](#basename)                                     |
+| `cat`              | [`File()`](#file) / [`Concat()`](#concat)                     |
+| `cut`              | [`Column()`](#column)                                         |
+| `dirname`          | [`Dirname()`](#dirname)                                       |
+| `echo`             | [`Echo()`](#echo)                                             |
+| `grep`             | [`Match()`](#match) / [`MatchRegexp()`](#matchregexp)         |
+| `grep -v`          | [`Reject()`](#reject) / [`RejectRegexp()`](#rejectregexp)     |
+| `head`             | [`First()`](#first)                                           |
+| `ls`               | [`ListFiles()`](#listfiles)                                   |
+| `sed`              | [`Replace()`](#replace) / [`ReplaceRegexp()`](#replaceregexp) |
+| `sha256sum`        | [`SHA256Sum()`](#sha256sum)                                   |
+| `tail`             | [`Last()`](#last)                                             |
+| `uniq -c`          | [`Freq()`](#freq)                                             |
+| `wc -l`            | [`CountLines()`](#countlines)                                 |
 
 # Sources, filters, and sinks
 
@@ -544,7 +544,7 @@ Each input file will be closed once it has been fully read.
 
 ## Dirname
 
-`Dirname()` reads a list of pathnames from the pipe, one per line, and returns a pipe which contains only the parent directories of each pathname (so, for example, `/usr/local/bin/foo` would become just `/usr/local/bin`). This is the complement of [Basename](#basename).
+`Dirname()` reads a list of file paths from the pipe, one per line, and returns a pipe which contains only the parent directories of each pathname (so, for example, `/usr/local/bin/foo` would become just `/usr/local/bin`). This is the complement of [Basename](#basename).
 
 If a line is empty, `Dirname()` will convert it to a single dot: `.` (this is the behaviour of Unix `dirname` and the Go standard library's `filepath.Dir`).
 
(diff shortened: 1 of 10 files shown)


bitfield:
Hey @thomaspoignant , awesome PR, thank you!

I wonder if there's also a case for treating the contents of the pipe as a list of files (as generated, for example, by `ListFiles()`), and transforming them into hash sums. That is to say, a SHA256 _filter_ as well as a _sink_, in `script` language. What do you think?

I also wonder if we should name this `SHA256`, to make it clear which algorithm is being used (there are a zillion different kinds of `CheckSum`, for example).

thomaspoignant:
Thanks @bitfield for you feedback.
I have rename the sink to `SHA256`.

Just to be sure, if we use `ListFiles()` and the result is something like 
``` 
testdata/multiple_files/1.txt
testdata/multiple_files/2.txt
testdata/multiple_files/3.tar.zip
```
What do you want as the result of the sink?
Something like:
```
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
b267dc7e66ee428bc8b51b1114bd0e05bde5c8c5d20ce828fbc95b83060c2f17
```

bitfield:
That looks pretty good, yes! In general the semantics seem to be that if you call a sink method (such as `String()`, you get a final result and an error, and that's the end of the pipe. That's appropriate for a single file, opened with `File()`, for example. On the other hand, if your pipe contains a list of _files_ (such as that produced by `ListFiles()`), then a filter method replaces each line with the transformed version of that line (like `DirName()`, for example).

I think it would be really useful if you could add an example program (in the `examples` directory) that shows the way someone might use this in a real-world context. For example, calculating hash sums of a bunch of files.

thomaspoignant:
@bitfield sorry to ask a lot of question 🤷‍♂️

But do you think I should keep a _filter_ and a _sink_?
If I want to keep them both I can have a name conflict because I have 2 function which are doing almost the same thing.

bitfield:
No need to apologise! I find discussions like this one quite useful for thrashing out how stuff should work and what makes sense to everybody.

I see your point about the naming conflict. I suppose it depends on what the example program looks like. Would we ever want to get the checksums for multiple files at once? If not, we don't need the filter, just the sink.

What do you think about `SHA256Sum()` as a name? Then if somebody wants `MD5Sum()` they could have that, and it matches the Unix command name.

By the way, one thing that occurred to me about the tests. We compare the result from `p.SHA256()` by checksumming the file ourselves using the `sha256` library... but that, of course, is how `p.SHA256()` is implemented! So we're not really testing anything here, except our ability to call a library. And if there should be a bug in that library, we won't know, will we?

A good practice with test expectations is to always calculate them yourself, by some other means, different to what you're using to implement the feature. So we could paste the contents of the test file into https://emn178.github.io/online-tools/sha256.html or some other tool, get the result as a hash, and compare that to `p.SHA256()` as a string literal.

thomaspoignant:
@bitfield I made some changes, SHA256Sum is now a filter and works after a `ListFiles()`.
I added an [example](https://github.com/bitfield/script/blob/e9043cd6e32d2a742500a2912bbe7e6e045b8c01/examples/sha256sum/main.go), I hope it is clearer than before what it do.

bitfield:
Looking good!

I feel like it would be good to have the ability to checksum both the contents of a pipe, as bytes, and a set of files. What do you think? If we did that then `SHA256Sum()` would work for the pipe, and maybe we could call the file version `SHA256Sums()`?

The example program is nice. How about we expand it slightly to compare the sums of two files, and tell you whether or not they're the same? I always like to make the example programs a tiny bit more interesting, if possible.

README.md
```diff
@@ -709,6 +711,17 @@ p := script.File("test.txt").Replace("old", "new")
 p := script.File("test.txt").ReplaceRegexp(regexp.MustCompile("Gol[a-z]{1}ng"), "Go")
 ```
 
+## SHA256Sum
+`SHA256Sum()` reads a list of path names from the pipe, one per line, and returns a pipe which contains the SHA-256 checksum of each pathname.
```
bitfield: How about `a list of file paths`? We can't checksum directories, I suppose.

README.md
```diff
@@ -709,6 +711,17 @@ p := script.File("test.txt").Replace("old", "new")
...
+If a line is empty, SHA256Sum will set an error to the pipe.
+
+Examples:
+
+| Input                             | `SHA256Sum` output                                                                                                                                                                                             |
+| --------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `testdata/sha256Sum.input.txt`    | `1870478d23b0b4db37735d917f4f0ff9393dd3e52d8b0efa852ab85536ddad8e`                                                                                                                                             |
+| `testdata/multiple_files/`        | `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`<br>`e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`<br>`e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` |
```
bitfield: I don't think this is right. If the line is a path to a directory, `SHA256Sum` will return an error, won't it? The input here is actually the set of three pathnames which represent the files in `multiple_files`.

testdata/sha256SumBin.input.txt
```diff
@@ -0,0 +1 @@
+other file
```
bitfield: I don't see where we use this?

README.md
thomaspoignant: Yes you are right, if this is a folder I will have an error, I will correct this example.

thomaspoignant:
@bitfield Thanks for your time.
I have create @ method now, one `filter` and one `sink`, can you tell me if it looks alright?

thomaspoignant:
@bitfield do you have any feedbacks on my updated pull request? 
Should I change something?

sinks.go
```diff
@@ -44,6 +46,25 @@ func (p *Pipe) CountLines() (int, error) {
 	return lines, p.Error()
 }
 
+// SHA256Sum create the SHA-256 of the file from the pipe's reader, and returns the
```
bitfield:
```suggestion
// SHA256Sum calculates the SHA-256 of the file from the pipe's reader, and returns the
```

sinks.go
```diff
@@ -44,6 +46,25 @@ func (p *Pipe) CountLines() (int, error) {
...
+
+	bytes, err := p.Bytes()
+	if err != nil {
+		p.SetError(err)
+		return "", p.Error()
+	}
+
+	sum := sha256.Sum256(bytes)
+	encodedCheckSum := hex.EncodeToString(sum[:])
+	return encodedCheckSum, p.Error()
```
bitfield:
```suggestion
	return encodedCheckSum, nil
```

sinks.go
bitfield: Methods in each of these files are kept in alphabetical order, just to make things a little saner; see the [contributor's guide](https://github.com/bitfield/script/blob/master/CONTRIBUTING.md#style-and-formatting).

sinks.go
bitfield:
```suggestion
	return encodedCheckSum, nil
```

sinks_test.go
```diff
@@ -84,6 +89,30 @@ func TestZeroPipeSinks(t *testing.T) {
 	doSinksOnPipe(t, &Pipe{}, "zero")
 }
 
+func TestSHA256Sums(t *testing.T) {
+	t.Parallel()
+
```
bitfield: I don't use any blank lines within functions in this project (I don't think), so for consistency of style, shall we remove them here?

sinks_test.go
```diff
@@ -84,6 +89,30 @@ func TestZeroPipeSinks(t *testing.T) {
 	doSinksOnPipe(t, &Pipe{}, "zero")
 }
 
+func TestSHA256Sums(t *testing.T) {
```
bitfield: Despite the name, this actually tests `SHA256Sum`, not `SHA256Sums`!

sinks_test.go
```diff
@@ -84,6 +89,30 @@ func TestZeroPipeSinks(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		p := File(tc.testFileName)
+		got, err := p.SHA256Sum()
+		if err != nil {
+			t.Error(err)
+		}
+		if got != tc.want {
+			t.Errorf("want %s checksum, got %x", tc.want[:], got[:])
```
bitfield:
```suggestion
			t.Errorf("want %q, got %q", tc.want, got)
```

sinks.go
thomaspoignant: The method just after is called `Stdout` so I think `SHA256` should be before this one.

thomaspoignant:
@bitfield I have tried to resolve all your feedbacks, and don't hesitate if you have more feedback.

examples/sha256sum/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+	"github.com/bitfield/script"
+	"log"
+	"os"
+	"strings"
+)
+
+func main() {
+	// Read the first 2 args, and calculate the checksum of the files.
+	hashFiles, err := script.Args().First(2).SHA256Sums().String()
```
bitfield, line 13: Slick!

examples/sha256sum/main.go
```diff
@@ -0,0 +1,27 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Compare the SHA256 checksum to check if files are identical.
+	checkSums := strings.Split(hashFiles,"\n")
```
bitfield, line 19: Seems to me we could use a `Slice()` sink, doesn't it? There are now several filters which produce a pipe full of newline-separated strings, and it would be convenient to get these right into a `[]string`.

examples/sha256sum/main.go
```diff
@@ -0,0 +1,27 @@
...
+	if checkSums[0] == checkSums[1]{
+		fmt.Println("Hashes are identical")
+		os.Exit(0)
+	} else {
```
bitfield, line 23: We can skip `else` here!

examples/visitors/go.sum
```diff
@@ -0,0 +1 @@
+github.com/bitfield/script v0.9.0/go.mod h1:MRqXqj+GhENKHXcjadvDTU5/Z1rgBHbtw/RSyHprCl4=
```
bitfield, line 1: We probably don't want to check in this `go.sum`.

examples/sha256sum/main.go
bitfield: (Not asking you to implement that; this is #45)

sinks.go
bitfield: Sorry, I misread this; ignore me!

examples/sha256sum/main.go
thomaspoignant: Yes, having a `Slice()` method could be awesome.

bitfield:
Great job, @thomaspoignant ! This is released as `v0.17.0`. Looking forward to your next contribution! Thanks again for all the hard work.

thomaspoignant:
Thank you @bitfield, I will check for the next contribution for sure.

Issue #45, opened 2020-04-02
Add 'Slice()' sink method to get pipe contents as a slice of string


bitfield:

thomaspoignant:
@bitfield I can do it if you want.
Just to be sure, this is just a split on the end of the line like this? https://github.com/bitfield/script/blob/34cdc5b9b45f9810cd0ce9ebf9e2a0f063ec271d/examples/sha256sum/main.go#L19

bitfield:
Something like that should work, yes!

thomaspoignant:
I have open a PR (#47) for that.

mrKappen:
hey @bitfield, I think this issue can be closed as the PR addressing it has already been merged

bitfield:
Thanks @mrKappen! You're quite right. My commit message at 9847d19e69d72d73b191e1f3c65d5dba94605a1c was wrong; it mentioned #47 instead of #45, so this issue wasn't auto-closed.

PR #47, opened 2020-04-06
Add Slice Sink

thomaspoignant:
It resolve this issue: #45

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@ examples/cat2/cat2
 examples/echo/echo
 examples/head/head
 examples/visitors/visitors
+examples/visitors/go.sum
 .vscode/settings.json
 examples/ls/ls
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,6 @@ examples/cat2/cat2
 examples/echo/echo
 examples/head/head
 examples/visitors/visitors
-examples/visitors/go.sum
+examples/*/go.sum
 .vscode/settings.json
 examples/ls/ls
--- a/README.md
+++ b/README.md
@@ -797,7 +797,7 @@ Unlike most sinks, `Read()` does not read the whole contents of the pipe (unless
 
 ## Slice
 
-`Slice()`, returns each line of the pipe as a Slice item, plus an error:
+`Slice()`, returns the contents of the pipe as a slice of strings, one element per line, plus an error:
 ```go
 var slice []string
 slice, err := script.Args().Slice()
--- a/README.md
+++ b/README.md
@@ -797,7 +797,7 @@ Unlike most sinks, `Read()` does not read the whole contents of the pipe (unless
 
 ## Slice
 
-`Slice()`, returns the contents of the pipe as a slice of strings, one element per line, plus an error:
+`Slice()` returns the contents of the pipe as a slice of strings, one element per line, plus an error:
 ```go
 var slice []string
 slice, err := script.Args().Slice()
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ John is also a [Kubernetes and cloud infrastructure consultant](https://bitfield
 	- [Bytes](#bytes)
 	- [CountLines](#countlines)
 	- [Read](#read)
+	- [Slice](#slice)
 	- [SHA256Sum](#sha256sum)
 		- [Why not MD5?](#why-not-md5)
 	- [Stdout](#stdout)
@@ -793,6 +794,14 @@ n, err := r.Read(buf)
 Because a Pipe is an `io.Reader`, you can use it anywhere you would use a file, network connection, and so on. You can pass it to `ioutil.ReadAll`, `io.Copy`, `json.NewDecoder`, and anything else which takes an `io.Reader`.
 
 Unlike most sinks, `Read()` does not read the whole contents of the pipe (unless the supplied buffer is big enough to hold them).
+
+## Slice
+
+`Slice()`, returns each line of the pipe as a Slice item, plus an error:
+```go
+var slice []string
+slice, err := script.Args().Slice()
+```
 
 ## SHA256Sum
 
--- a/README.md
+++ b/README.md
@@ -119,9 +119,9 @@ John is also a [Kubernetes and cloud infrastructure consultant](https://bitfield
 	- [Bytes](#bytes)
 	- [CountLines](#countlines)
 	- [Read](#read)
-	- [Slice](#slice)
 	- [SHA256Sum](#sha256sum)
 		- [Why not MD5?](#why-not-md5)
+	- [Slice](#slice)
 	- [Stdout](#stdout)
 	- [String](#string)
 	- [WriteFile](#writefile)
@@ -795,14 +795,6 @@ Because a Pipe is an `io.Reader`, you can use it anywhere you would use a file, 
 
 Unlike most sinks, `Read()` does not read the whole contents of the pipe (unless the supplied buffer is big enough to hold them).
 
-## Slice
-
-`Slice()` returns the contents of the pipe as a slice of strings, one element per line, plus an error:
-```go
-var slice []string
-slice, err := script.Args().Slice()
-```
-
 ## SHA256Sum
 
 `SHA256Sum()`, as the name suggests, returns the [SHA256 checksum](https://en.wikipedia.org/wiki/SHA-2) of the file as a hexadecimal number stored in a string, plus an error:
@@ -813,6 +805,14 @@ sha256Sum, err := script.File("test.txt").SHA256Sum()
 ### Why not MD5?
 
 [MD5 is insecure](https://en.wikipedia.org/wiki/MD5#Security).
+
+## Slice
+
+`Slice()` returns the contents of the pipe as a slice of strings, one element per line, plus an error:
+```go
+var slice []string
+slice, err := script.Args().Slice()
+```
 
 ## Stdout
 
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ John is also a [Kubernetes and cloud infrastructure consultant](https://bitfield
 	- [Read](#read)
 	- [SHA256Sum](#sha256sum)
 		- [Why not MD5?](#why-not-md5)
-	- [Slice](#slice)
+	- [Slice](#slice-1)
 	- [Stdout](#stdout)
 	- [String](#string)
 	- [WriteFile](#writefile)
@@ -453,13 +453,13 @@ fmt.Println(files)
 `Slice()` creates a pipe from a slice of strings, one per line.
 
 ```go
-p := Slice([]string{"1", "2", "3"})
+p := script.Slice([]string{"1", "2", "3"})
 output, err := p.String()
 fmt.Println(output)
 // Output:
-1
-2
-3
+// 1
+// 2
+// 3
 ```
 
 ## Stdin
@@ -808,10 +808,13 @@ sha256Sum, err := script.File("test.txt").SHA256Sum()
 
 ## Slice
 
-`Slice()` returns the contents of the pipe as a slice of strings, one element per line, plus an error:
+`Slice()` returns the contents of the pipe as a slice of strings, one element per line, plus an error. An empty pipe will produce an empty slice. A pipe containing a single empty line (that is, a single `\n` character) will produce a slice of one element which is the empty string.
+
 ```go
-var slice []string
-slice, err := script.Args().Slice()
+args, err := script.Args().Slice()
+for _, a := range args {
+	fmt.Println(a)
+}
 ```
 
 ## Stdout
@@ -880,6 +883,7 @@ Since `script` is designed to help you write system administration programs, a f
 * [least_freq](examples/least_freq/main.go)
 * [ls](examples/ls/main.go)
 * [sha256sum](examples/sha256sum/main.go)
+* [slice](examples/slice/main.go)
 * [tail](examples/tail/main.go)
 * [visitors](examples/visitors/main.go)
 
--- a/examples/slice/main.go
+++ b/examples/slice/main.go
@@ -7,14 +7,12 @@ import (
 )
 
 func main() {
-	// Get a Slice with all the running process PID
-	pids, err := script.Exec("ps -ea").Column(1).Slice()
+	errors, err := script.Args().Concat().Match("Error").First(10).Slice()
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	fmt.Printf("Running process PIDs are:")
-	for _, pid := range pids {
-		fmt.Println(pid)
+	for _, e := range errors {
+		fmt.Printf("ERROR: %s", e)
 	}
 }
--- a/examples/slice/main.go
+++ b/examples/slice/main.go
@@ -14,7 +14,7 @@ func main() {
 	}
 
 	fmt.Printf("Running process PIDs are:")
-	for _, pid := range pids{
+	for _, pid := range pids {
 		fmt.Println(pid)
 	}
 }
--- a/examples/slice/main.go
+++ b/examples/slice/main.go
@@ -7,11 +7,11 @@ import (
 )
 
 func main() {
-	// Read args, and return it in a Slice.
-	params, err := script.Args().Slice()
+	// Get a Slice with all the running process PID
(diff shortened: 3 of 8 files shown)


README.md
```diff
@@ -794,6 +795,14 @@ Because a Pipe is an `io.Reader`, you can use it anywhere you would use a file,
 
 Unlike most sinks, `Read()` does not read the whole contents of the pipe (unless the supplied buffer is big enough to hold them).
 
+## Slice
+
+`Slice()`, returns each line of the pipe as a Slice item, plus an error:
```
bitfield, line 800:
```suggestion
`Slice()`, returns the contents of the pipe as a slice of strings, one element per line, plus an error:
```

examples/slice/main.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"fmt"
+	"github.com/bitfield/script"
+	"log"
+)
+
+func main() {
+	// Read args, and return it in a Slice.
+	params, err := script.Args().Slice()
```
bitfield, line 11: Can we make this example program a bit more interesting? Ideally, each one would use `script` to do something potentially useful, or be a toy example of something useful.

examples/visitors/go.sum
```diff
@@ -0,0 +1 @@
+github.com/bitfield/script v0.9.0/go.mod h1:MRqXqj+GhENKHXcjadvDTU5/Z1rgBHbtw/RSyHprCl4=
```
bitfield, line 1: We probably don't want to check in the `go.sum` for the example programs.

sinks.go
```diff
@@ -46,6 +46,20 @@ func (p *Pipe) CountLines() (int, error) {
...
 }
 
+// Slice return each lines of the pipe in a Slice or an error.
+// If there is an error reading the pipe, the pipe's error status
+// is also set.
+func (p *Pipe) Slice() ([]string, error) {
+	if p == nil || p.Error() != nil {
+		return nil, p.Error()
+	}
+	var slice []string
```
bitfield, line 56: We don't usually name variables after their type, as tempting as it is sometimes. How about `result` here instead?

sinks.go
```diff
@@ -46,6 +46,20 @@ func (p *Pipe) CountLines() (int, error) {
...
+	p.EachLine(func(line string, out *strings.Builder) {
+		slice = append(slice, line)
```
bitfield, line 58: This is probably a bit inefficient; how about getting the result of `p.String()` instead and splitting on newlines? Don't forget to make sure the last line also ends with a newline, if the contents of the pipe does.

sinks_test.go
```diff
@@ -166,6 +172,30 @@ func TestSHA256Sum(t *testing.T) {
 	}
 }
 
+func TestSinkSlice(t *testing.T) {
```
bitfield, line 175:
```suggestion
func TestSliceSink(t *testing.T) {
```

sinks_test.go
```diff
@@ -166,6 +172,30 @@ func TestSHA256Sum(t *testing.T) {
...
+	t.Parallel()
+
+	want := []string {
+		"testdata/multiple_files/1.txt",
+		"testdata/multiple_files/2.txt",
+		"testdata/multiple_files/3.tar.zip",
+	}
+	got, err := ListFiles("testdata/multiple_files").Slice()
```
bitfield, line 183: Aren't we testing ListFiles here too, which we don't need to do? I would use a test case of a string containing embedded newlines, and test that we get them correctly as a slice.

sinks_test.go
```diff
@@ -166,6 +172,30 @@ func TestSHA256Sum(t *testing.T) {
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	if !reflect.DeepEqual(got, want) {
```
bitfield, line 188: I suggest using `go-cmp` here (see `TestSlice` for an example), because the diff output is much better.

sinks_test.go
```diff
@@ -166,6 +172,30 @@ func TestSHA256Sum(t *testing.T) {
...
+		t.Errorf("want %q, got %q", want, got)
+	}
+
+	// Empty pipe, should return empty slice
+	got, _ = File("testdata/empty.txt").Slice()
```
bitfield, line 193: We should receive and check the error value here, too.

examples/visitors/go.sum
thomaspoignant: This is the second time I made the mistake, do you mind if I put it in the gitignore?

sinks.go
thomaspoignant: I am not sure of what to do when the pipe is empty.
Should we have a `Slice with 1 empty element` or an `empty Slice`?

And just to be sure, if the pipe ends with a new line, the last element of the Slice will be an empty String, right?

examples/visitors/go.sum
bitfield: Go ahead! That's a good idea.

sinks.go
bitfield: Actually, when the pipe is empty, there are _no lines_, so I think it makes sense to return a slice with _no elements_, despite my previous comment.

If the pipe consists of a single newline, then the result of Slice() should be a slice with 1 element which is a string containing a newline, I think.

examples/slice/main.go
thomaspoignant: I tried to have a better example let me know if it sounds good to you.

thomaspoignant:
@bitfield I have made the requested changes.
I hope this is what you have in mind?

.gitignore
```diff
@@ -4,5 +4,6 @@ examples/cat2/cat2
 examples/echo/echo
 examples/head/head
 examples/visitors/visitors
+examples/visitors/go.sum
```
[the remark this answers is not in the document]
bitfield: It's not really relevant to your PR, but I think this is worth doing, isn't it?

README.md
```diff
@@ -794,6 +795,14 @@ Because a Pipe is an `io.Reader`, you can use it anywhere you would use a file,
 
 Unlike most sinks, `Read()` does not read the whole contents of the pipe (unless the supplied buffer is big enough to hold them).
 
+## Slice
+
+`Slice()`, returns the contents of the pipe as a slice of strings, one element per line, plus an error:
```
bitfield, line 800:
```suggestion
`Slice()` returns the contents of the pipe as a slice of strings, one element per line, plus an error:
```

examples/slice/main.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func main() {
+	// Get a Slice with all the running process PID
+	pids, err := script.Exec("ps -ea").Column(1).Slice()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Printf("Running process PIDs are:  %q", pids)
```
bitfield, line 16: Okay, but this doesn't really make use of them as a _slice_, does it? If we want to demonstrate how useful it is to turn a pipe into a slice, we could `range` over that slice, for example.

sinks.go
```diff
@@ -46,6 +46,32 @@ func (p *Pipe) CountLines() (int, error) {
 	return lines, p.Error()
 }
 
+// Slice return each lines of the pipe in a Slice or an error.
```
bitfield, line 49:
```suggestion
// Slice returns the contents of the pipe as a slice of strings, one element per line, or an error.
```

sinks.go
```diff
@@ -46,6 +46,32 @@ func (p *Pipe) CountLines() (int, error) {
...
+// If there is an error reading the pipe, the pipe's error status
+// is also set.
+func (p *Pipe) Slice() ([]string, error) {
+	if p == nil || p.Error() != nil {
+		return nil, p.Error()
+	}
+	value, err := p.String()
```
bitfield, line 56: It's a shame to allocate memory for the entire contents of the pipe as a string, and then throw it away, isn't it? Maybe it's a better idea to read the pipe using `EachLine()`, and build up the slice of strings that way?

sinks.go
```diff
@@ -46,6 +46,32 @@ func (p *Pipe) CountLines() (int, error) {
 	return lines, p.Error()
 }
 
+// Slice return each lines of the pipe in a Slice or an error.
+// If there is an error reading the pipe, the pipe's error status
+// is also set.
+func (p *Pipe) Slice() ([]string, error) {
```
bitfield, line 52: I think this method is out of order; it should go after `SHA256Sum`.

bitfield:
Hi @thomaspoignant, how are you getting on? Would you like any help with this? Are you ready for another review yet?

thomaspoignant:
Hi @bitfield sorry I was not able to work on this, I will try to work on this next week.

I am really sorry about that delay.

bitfield:
Don't worry about it! I don't get nearly as much time as I'd like to work on this project either. If you'd like some help finishing this off, just let me know. Otherwise, there's no hurry.

sinks.go
thomaspoignant: That's what I did in my first version : https://github.com/bitfield/script/blob/f74894fd9e143fecf4195bd266b091e766370086/sinks.go#L57 
but you had this [feedback](https://github.com/bitfield/script/pull/47/commits/f74894fd9e143fecf4195bd266b091e766370086#r407059026) that's why I end up with this solution.

thomaspoignant:
@bitfield I have made almost all the changes you requested.
The last thing is how we iterate over the string, it is not clear to me if we should use `String()` or `EachLine()` to iterate.

bitfield:
If we can use EachLine(), that's probably more efficient.

thomaspoignant:
@bitfield Should be good now, it use the version with `EachLine()`.

README.md
```diff
@@ -119,6 +119,7 @@ John is also a [Kubernetes and cloud infrastructure consultant](https://bitfield
 	- [Bytes](#bytes)
 	- [CountLines](#countlines)
 	- [Read](#read)
+	- [Slice](#slice)
```
bitfield, line 122: This should come after SHA256Sum() if they're in alphabetical order, shouldn't it?

README.md
```diff
@@ -794,6 +795,14 @@ Because a Pipe is an `io.Reader`, you can use it anywhere you would use a file,
 
 Unlike most sinks, `Read()` does not read the whole contents of the pipe (unless the supplied buffer is big enough to hold them).
 
+## Slice
```
bitfield, line 798: Again, let's move this after the documentation of SHA256Sum().

examples/slice/main.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"fmt"
+	"github.com/bitfield/script"
+	"log"
+)
+
+func main() {
+	// Get a Slice with all the running process PID
+	pids, err := script.Exec("ps -ea").Column(1).Slice()
```
bitfield, line 11: Yes! This is a great use case for Slice(), exactly what we need.

However, wherever possible I try to avoid calling userland tools like `ps` in example programs. After all, one of the big applications of the `script` library is in scratch containers and other places where we can't rely on having any of the standard Unix tools installed. 

Suppose we did something like in the introductory README example:

```go
script.Args().Concat().Match("Error").First(10).Stdout()
```

Only instead of sending the error matches to `Stdout()`, we range over them as a slice and print them out:

```go
for _, e := range errors {
	fmt.Printf("ERROR: %s", e)
}
```

pipes_test.go
```diff
@@ -107,6 +107,8 @@ func doMethodsOnPipe(t *testing.T, p *Pipe, kind string) {
 	p.Concat()
 	action = "CountLines()"
 	p.CountLines()
+	action = "Slice()"
```
bitfield, line 110: Same ordering issue here.

sinks.go
```diff
@@ -65,6 +65,20 @@ func (p *Pipe) SHA256Sum() (string, error) {
 	return encodedCheckSum, nil
 }
 
+// Slice returns the contents of the pipe as a slice of strings, one element per line, or an error.
+// If there is an error reading the pipe, the pipe's error status
+// is also set.
```
bitfield, line 70: Looks like line wrapping went a bit wrong here—this third line would fit on the line above, wouldn't it?

sinks_test.go
```diff
@@ -166,6 +172,60 @@ func TestSHA256Sum(t *testing.T) {
...
+		"testdata/multiple_files/1.txt",
+		"testdata/multiple_files/2.txt",
+		"testdata/multiple_files/3.tar.zip",
+	}
+	got, err := input.Slice()
+	if err != nil {
+		t.Error(err)
+	}
+
+	if !cmp.Equal(got, want) {
```
bitfield, line 189: It's a small point, but I always try to write `cmp.Equal(want, got)`, because then they follow in the same order on the next line: `cmp.Diff(want, got)`. If you diff the results in the opposite order, they look wrong, so I use this order for both the comparison and the diff.

sinks_test.go
```diff
@@ -166,6 +172,60 @@ func TestSHA256Sum(t *testing.T) {
...
+		t.Error(cmp.Diff(want, got))
+	}
+
+	// Empty pipe, should return empty slice
+	got, err = Echo("").Slice()
+	if err != nil {
+		t.Error(err)
+	}
+	if len(got) != 0 {
+		t.Errorf("want %d, got %d", 1, len(got))
```
bitfield, line 199:
```suggestion
		t.Errorf("want zero-length slice, got %v", got)
```

sinks_test.go
```diff
@@ -166,6 +172,60 @@ func TestSHA256Sum(t *testing.T) {
...
+	want = []string{""}
+	got, err = Echo("\n").Slice()
+	if err != nil {
+		t.Error(err)
+	}
+	if !cmp.Equal(want, got) {
+		t.Error(cmp.Diff(want, got))
+	}
+
+	// Empty line between two existing line
```
bitfield, line 212:
```suggestion
	// Empty line between two existing lines
```

examples/slice/main.go
thomaspoignant: Yes you are right this is a way better example.

thomaspoignant:
Sorry about all these order problems, it should be fine now.

thomaspoignant:
@bitfield do you need more work on this PR?

examples/slice/main.go
```diff
@@ -0,0 +1,18 @@
+package main
```
bitfield, line 1: Nice! Let's add a comment or two explaining what this program does, and ideally let's supply an example text file with it which contains suitable matching lines, and explain how to run the example on this data.

sinks_test.go
```diff
@@ -166,6 +172,60 @@ func TestSHA256Sum(t *testing.T) {
 	}
 }
 
+func TestSliceSink(t *testing.T) {
+	t.Parallel()
+	input := Echo("testdata/multiple_files/1.txt\ntestdata/multiple_files/2.txt\ntestdata/multiple_files/3.tar.zip\n")
```
bitfield, line 177: Can you turn this into a table-driven test, with named test cases, using `t.Run()`?

thomaspoignant:
I send the changes if you could have a look.

sinks_test.go
```diff
@@ -166,6 +172,61 @@ func TestSHA256Sum(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			p := tt.fields
+			got, err := p.Slice()
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Slice() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !cmp.Equal(got, tt.want) {
```
bitfield, line 223: I like to always compare in the order `want, got`, because when I print the diff, that's the order that makes sense. And though it makes no difference to Go, it's pleasing to me to have them in the same order everywhere!

examples/slice/main.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"fmt"
+	"github.com/bitfield/script"
+	"log"
+)
+
+func main() {
+	// Read the first args, and put the 10 first lines which match the word Error.
```
bitfield, line 10: Great! Do we have an example input file for this?

thomaspoignant:
There is an example file + I have changed the order in comparison.

bitfield:
Great job @thomaspoignant! Thanks a lot for the help. I've given it a final pass for polish, and merged this as `v0.18.0`. Enjoy!

Issue #51, opened 2020-04-26
Get user input


bitfield:
It's common when writing scripts that install or configure things to need input from the user (at a minimum, something like 'Press Enter to continue'; at a maximum, to be able to prompt the user for input, with an optional default value, and return that value).

Let's use this issue to design how that would look, and I invite suggestions!

mathantunes:
Hi @bitfield ,

I'm quite new to the Open Source Community (read this is my first try) but I have some experience with Go.

So I sketched out a couple of the possibilities here.
You mentioned installation scripts and configuration, I also thought basic CLI menus should be included here.

So here are my thoughts, please feel totally free to change anything or everything.

# Specification on Reading User input for Script 

## Input Types

### Read from a user input

* Read String (e.g. What's your name ?)
* Read Number (e.g. How old are you ?)
* Read Array (e.g. List car brands you like)

### Protected

* Read Password

### Menu Items

* Read from a Menu

Tell a user to pick from a certain set of options
A menu can be defined as an array of choices

### Optional parameters

* A message echoed to the user, on the same line as
	What is your name? _______

* A default value
	Would you like to continue [Y/n] default(Y):

* An StdErr message and end of command execution for non-opntional reads
	When a user types a string instead of number, Pipe.WithError

```go
type Options struct {
	Required bool
	Default interface{}
	OnError func(in interface{}) *Pipe //WithError
}

opts := *&Options{
	Required : true,
	Default : nil,
}

vStr := p.ReadString("Input your name:", opts) //string
vFlt := p.ReadNumber("Input your age:", opts) //float64
vPwd := p.ReadPassword("Password: ", opts) //string

(p *Pipe) ReadString(message string, opts Options) *Pipe {
//Execute read
//Return New Pipe
}

vArr := p.ReadArray("Input cars you like", opts)

items := []string{"1 - Skip", "2 - Continue", "3 - Quit"}
(p *Pipe) DisplayMenu(items []string, opts Options, f func(chosen string)*Pipe) *Pipe {
//Display List
//Read user input
//Call f with item picked
//Return new pipe created on f
})
```

bitfield:
Hey @manthantunes, great suggestions, thank you! This looks like a good starting point.

Let's see if we can put together an example program which uses the input feature for some real-world application. (This is often a great way to start designing something: use it!)

For example, a software installer which needs to get the installation directory from the user.

mathantunes:
Hi @bitfield,

[...]

PR #52, opened 2020-05-02
#51 sources/prompt - Implement Get User Input

mathantunes:
The implementation of ```Prompt``` function that reads user input or uses a default value provided on the function parameters.

Unit tests, README example and example on ```examples/``` folder

Issue #51

### DIFF
--- a/examples/prompt/main.go
+++ b/examples/prompt/main.go
@@ -1,9 +1,20 @@
 package main
 
 import (
+	"strings"
+
 	"github.com/bitfield/script"
 )
 
 func main() {
-	script.Prompt("This is an example, what do you think? ", "It's awesome").Stdout()
+	script.Echo("Installation of Fake Product\n").Stdout()
+	curDir, err := script.Exec("pwd").String()
+	if err != nil {
+		panic(err)
+	}
+
+	script.Prompt("Choose install location: ", strings.TrimSpace(curDir)).
+		ExecForEach("touch {{.}}/fake_install.sh").Stdout()
+
+	script.Echo("Installation of Fake Product complete").Stdout()
 }
--- a/examples/prompt/main.go
+++ b/examples/prompt/main.go
@@ -1,0 +1,9 @@
+package main
+
+import (
+	"github.com/bitfield/script"
+)
+
+func main() {
+	script.Prompt("This is an example, what do you think? ", "It's awesome").Stdout()
+}
--- a/examples/prompt/main.go
+++ b/examples/prompt/main.go
@@ -13,7 +13,7 @@ func main() {
 		panic(err)
 	}
 
-	script.Prompt("Choose install location: ", strings.TrimSpace(curDir)).
+	script.Prompt("Choose install location", strings.TrimSpace(curDir)).
 		ExecForEach("touch {{.}}/fake_install.sh").Stdout()
 
 	script.Echo("Installation of Fake Product complete").Stdout()
--- a/sources.go
+++ b/sources.go
@@ -1,7 +1,6 @@
 package script
 
 import (
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -117,9 +116,10 @@ func Stdin() *Pipe {
 
 // Prompt presents a message to the user and returns a pipe containing user input or defaultValue provided.
 func Prompt(message, defaultValue string) *Pipe {
-	fmt.Print(message)
 
-	in, err := Stdin().First(1).String()
+	Echo(message).Stdout()
+	p := Stdin().First(1)
+	in, err := p.String()
 	if err != nil {
 		return NewPipe().WithError(err)
 	}
@@ -127,5 +127,5 @@ func Prompt(message, defaultValue string) *Pipe {
 	if in == "" {
 		in = defaultValue
 	}
-	return NewPipe().WithReader(strings.NewReader(in))
+	return Echo(in)
 }
--- a/sources.go
+++ b/sources.go
@@ -116,7 +116,6 @@ func Stdin() *Pipe {
 
 // Prompt presents a message to the user and returns a pipe containing user input or defaultValue provided.
 func Prompt(message, defaultValue string) *Pipe {
-
 	Echo(message).Stdout()
 	p := Stdin().First(1)
 	in, err := p.String()
--- a/sources.go
+++ b/sources.go
@@ -1,6 +1,7 @@
 package script
 
 import (
+	"bufio"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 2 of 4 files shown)


examples/prompt/main.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+import (
+	"github.com/bitfield/script"
+)
+
+func main() {
+	script.Prompt("This is an example, what do you think? ", "It's awesome").Stdout()
```
bitfield, line 8: Yes, it's awesome, but I have a rule of thumb for example programs in this repo which says that they should not just demonstrate the feature, but actually show why it's useful. Why is it useful to be able to get input from the user? What real-world problem might we be solving by doing this? Let's write _that_ program.

This is helpful not only because it gives users something they can copy and modify for their own programs, but it also forces us to do some thinking. _Is_ the feature really useful, and if so, for what? That helps us design the feature.

sources.go
```diff
@@ -113,3 +113,19 @@ func Slice(s []string) *Pipe {
 func Stdin() *Pipe {
 	return NewPipe().WithReader(os.Stdin)
 }
+
+// Prompt presents a message to the user and returns a pipe containing user input or defaultValue provided.
+func Prompt(message, defaultValue string) *Pipe {
+
```
bitfield, line 119: It's a house style thing in this project that we don't use any blank lines within functions.

sources.go
```diff
@@ -113,3 +113,19 @@ func Slice(s []string) *Pipe {
 func Stdin() *Pipe {
 	return NewPipe().WithReader(os.Stdin)
 }
+
+// Prompt presents a message to the user and returns a pipe containing user input or defaultValue provided.
+func Prompt(message, defaultValue string) *Pipe {
```
bitfield, line 118: To make it easier to find things, functions are listed in alphabetical order in the source files (see the [Contributing](https://github.com/bitfield/script/blob/master/CONTRIBUTING.md) guide for details).

sources.go
```diff
@@ -113,3 +113,19 @@ func Slice(s []string) *Pipe {
...
+
+	Echo(message).Stdout()
+	p := Stdin().First(1)
```
bitfield, line 121: Should we show a prompt character, perhaps, to indicate to the user that the program is waiting for input?

bitfield: We should indicate the default value to the user too, shouldn't we?

sources.go
```diff
@@ -113,3 +113,19 @@ func Slice(s []string) *Pipe {
...
+	in, err := p.String()
+	if err != nil {
+		return NewPipe().WithError(err)
+	}
+	in = strings.TrimSpace(in)
```
bitfield, line 126: Do we want to silently remove leading and trailing spaces? I can imagine situations where they might be significant.

sources.go
```diff
@@ -113,3 +113,19 @@ func Slice(s []string) *Pipe {
...
+	if in == "" {
+		in = defaultValue
```
bitfield, line 128:
```suggestion
		return Echo(defaultValue)
```

sources_test.go
```diff
@@ -254,3 +254,49 @@ func TestStdin(t *testing.T) {
...
+
+func TestPrompt(t *testing.T) {
+	tcs := []struct {
+		Name        string
+		Input       string
+		Default     string
+		Want        string
+		ErrExpected bool
+	}{
+		{"User Input OK", "123", "notused", "123", false},
```
bitfield, line 266: It's not immediately clear with field-less literals like this which field is which. Let's write these as multi-line literals with field names, to make it clear.

sources_test.go
```diff
@@ -254,3 +254,49 @@ func TestStdin(t *testing.T) {
...
+		{"No User Input", "", "default", "default", false},
+		{"Complex Input", "/user/test @@@ testing one two three", "default", "/user/test @@@ testing one two three", false},
+	}
+	for _, tc := range tcs {
+		t.Run(tc.Name, func(t *testing.T) {
+			want := []byte(tc.Input)
```
bitfield, line 272: Actually, the output is going to be a string, isn't it? So the `want` value should be a string too.

sources_test.go
```diff
@@ -254,3 +254,49 @@ func TestStdin(t *testing.T) {
...
+			}
+
+			if _, err := tmp.Seek(0, 0); err != nil {
+				t.Error(err)
+			}
+
+			// Restore original Stdin.
+			oldStdin := os.Stdin
+			defer func() { os.Stdin = oldStdin }()
+			os.Stdin = tmp
```
bitfield, line 291: Not concurrency-safe; see `TestStdin` for a safe way to do this.

examples/prompt/main.go
mathantunes: I have updated the example with our installation example discussed on the issue #51

sources.go
mathantunes: Removed blank lines

sources.go
mathantunes: Corrected

sources.go
mathantunes: The message now displays the Prompt message and the default value between [ ].

What do you mean by prompt character ? Something like a blinking text input ?

sources.go
mathantunes: I have done it because using the 
```Stdin().First(1)```
To read user input adds a '\n` character at the end of the input.

```go
//On First
output.WriteString(scanner.Text())
output.WriteRune('\n')
```

A solution would be to utilize a different Stdin read method on ```fmt.Scan``` or ```bufio.Read```

sources.go
mathantunes: Done

sources.go
bitfield: That looks good! We should maybe leave a space after the colon before the user's input?

mathantunes: Done

sources.go
mathantunes: I have changed the implementation of user input.

Instead of having ```Stdin().First(1)```, the code now scans the ```Stdin()``` pipe

sources_test.go
mathantunes: I am having a little trouble on the implementation of this test.

When utilizing the ```exec.Command``` method, with a new **SCRIPT_TEST** defined for ```prompt``` as 
```go
case "prompt":
		Prompt("Input", "Default").Stdout()
```

The ```cmd.Output()``` is actually the first Stdout message from ```Prompt```, which is the message to the user ```Echo(fmt.Sprintf("%s [%s]: ", message, defaultValue)).Stdout()```

sources_test.go
mathantunes: Done

bitfield:
@mathantunes, how are you getting on? Are you ready for another review yet?

Issue #55, opened 2020-05-24
Have FindFiles() take a glob


bitfield:
For symmetry with ListFiles().

vigneshashokan:
Hey @bitfield ! Sorry for the delayed response. Before I start work on this, I wanted to clarify a thing.

Directory structure
```
root
├─── root_folder1
│    ├─── root_folder1_file1.txt
│    ├─── root_folder1_file2.txt
│    ├─── root_folder1_sub_folder1
│    │    ├── root_folder1_sub_folder1_file1.txt
│    │    └── root_folder1_sub_folder1_file2.txt
│    └─── root_folder1_sub_folder2
│         ├── root_folder1_sub_folder2_file1.txt
│         └── root_folder1_sub_folder2_file2.txt
└─── root_folder2
```

What is the expected output when I do a FindFiles("/root/*/root_folder1_sub*")?
My understanding is that the glob should be resolved to the actual folder paths, in this case [/root/root_folder1/root_folder1_sub_folder1, /root/root_folder1/root_folder1_sub_folder2] and list all the files under these folders
```
Output:
/root/root_folder1/root_folder1_sub_folder1/root_folder1_sub_folder1_file1.txt
/root/root_folder1/root_folder1_sub_folder1/root_folder1_sub_folder1_file2.txt
/root/root_folder1/root_folder1_sub_folder2/root_folder1_sub_folder2_file1.txt
/root/root_folder1/root_folder1_sub_folder2/root_folder1_sub_folder2_file2.txt
```
so the files /root/root_folder1/root_folder1_file1.txt and /root/root_folder1/root_folder1_file2.txt are not printed.

Is my understanding correct?

(sorry for the long file names)

bitfield:
I'm not quite sure what you're asking.

vigneshashokan:
@bitfield I wanted to understand the issue better. When you say that `FindFiles()` should be able to take a glob, that means,
1. It needs to take an input with wildcard chars like filepath = `/folder1/folder*`, and should list all the files recursively in this path.
2. It needs to take an input without wildcard chars as well like filepath = `/folder1/folder2`, and should list all the files recursively in this path.
Is this correct?

bitfield:
Correct! Basically, it should be the equivalent of ListFiles with a glob, followed by a FindFiles on all its results! (Maybe we could use this as a test.)

vigneshashokan:
Sweet that clears things up, Thanks!

vigneshashokan:
[...]

PR #58, opened 2020-06-10
#55 - impleneted glob for FindFiles()

prepodavan:
Implemented glob for `FindFiles()`


Prepodavan:
Yes, i did. I misinterpreted task. My mistake. So issue is not relevant anymore. Just nvm, everything is ok with this way of glob recognizing. 
Have u smth to say about this pr?

Issue #62, opened 2020-09-08
FindFiles and ListFiles might want to trim newlines from the path


christophberger:
Hi @bitfield , thanks for this awesome package. I just ran into an issue that is surely an easy thing to improve.

Issue: the output of Pipe.String() may include a newline. FindFiles and ListFiles cannot find a path that ends in a newline. Hence these two methods cannot consume the output of Pipe.String() without prior removal of newlines.

Suggestion: Have FindFiles and ListFiles trim trailing newlines from the input path.

Example:
```go 
package main

import (
	"log"
	"github.com/bitfield/script"
)

func main() {
	r, _ := script.Args().String()
	log.Printf("args: [%s]\n", r)
	p := script.ListFiles(r)
	_, err := p.Stdout()
	log.Println(err)
}
```
Call:
```
go run main.go .
```
Output: 
```
2020/09/08 13:57:58 args: [.
]
2020/09/08 13:57:58 stat .
: no such file or directory
```
Severity: minor annoyance but can cost time to track down.

bitfield:
Hi @christophberger, nice suggestion, thank you! It seems like there are actually a bunch of pipe sources which could struggle with a newline-terminated string. Maybe another solution is to provide a way to trim the trailing newline from the pipe's output? Ruby and Perl call this `chomp()`, I think. We could do the same. Ideas welcome!

christophberger:
I think you can best decide which approach fits best into the pipe design. My own bias would be towards automatic stripping of newlines from path strings. After all, a file path never contains newlines.

The reason for this bias is that I needed a while to figure out why FindFiles or ListFiles did not seem to work the way they should. Only after adding all the missing error handling and logging, I discovered the trailing newline as the reason for the malfunction. (Yeah, PEBCAK definitely. Lesson learned: Never neglect proper error handling, even for simple throw-away scripts.) If the newlines are stripped automatically, user errors like this one cannot happen anymore.

bitfield:
> Lesson learned: Never neglect proper error handling, even for simple throw-away scripts

As a Go teacher, I naturally agree with this 😁 	

> My own bias would be towards automatic stripping of newlines from path strings. After all, a file path never contains newlines.

That is a very good point, so any method which accepts file paths could trim them automatically.

Are you interested in working on this? If not, no problem, but if you are, we could work together on a PR for the fix.

christophberger:
Hi John,

[...]

PR #63, opened 2020-09-11
Trim newlines from file paths

christophberger:
Hi @bitfield , as discussed, I created a fix for issue #62 . 
Improvement: all functions taking a file path as input trim any trailing newline off the input string. This helps avoiding errors or the need of manual newline removal when the file path is a result of ListFiles() or FindFiles() (that always return a newline-delimited string).
I changed existing tests to receive paths with a trailing newline. 

Possible points to discuss:
* Instead of changing the existing test cases, separate test cases could be added to test with and without trailing newlines. On the other hand, the "without trailing newlines" case is already covered, as a newline is stripped anyway.
* Instead of just trimming trailing newlines, would trimming all leading and trailing whitespace from file paths make sense? Would there be a use case for that?

### DIFF
--- a/sources.go
+++ b/sources.go
@@ -34,6 +34,7 @@ func Exec(s string) *Pipe {
 // status will be set, and if the file does exist, the pipe will have no error
 // status.
 func IfExists(filename string) *Pipe {
+	filename = strings.TrimRight(filename, "\n")
 	_, err := os.Stat(filename)
 	if err != nil {
 		return NewPipe().WithError(err)
@@ -45,6 +46,7 @@ func IfExists(filename string) *Pipe {
 // starting pipelines. If there is an error opening the file, the pipe's error
 // status will be set.
 func File(name string) *Pipe {
+	name = strings.TrimRight(name, "\n")
 	p := NewPipe()
 	f, err := os.Open(name)
 	if err != nil {
@@ -58,6 +60,7 @@ func File(name string) *Pipe {
 // `find -type f`. If the path doesn't exist or can't be read, the pipe's error
 // status will be set.
 func FindFiles(path string) *Pipe {
+	path = strings.TrimRight(path, "\n")
 	var fileNames []string
 	walkFn := func(path string, info os.FileInfo, err error) error {
 		if err != nil {
@@ -78,6 +81,7 @@ func FindFiles(path string) *Pipe {
 // supplied path, one per line. The path may be a glob, conforming to
 // filepath.Match syntax.
 func ListFiles(path string) *Pipe {
+	path = strings.TrimRight(path, "\n")
 	if strings.ContainsAny(path, "[]^*?\\{}!") {
 		fileNames, err := filepath.Glob(path)
 		if err != nil {
--- a/sources_test.go
+++ b/sources_test.go
@@ -111,7 +111,7 @@ func TestFile(t *testing.T) {
 	t.Parallel()
 	wantRaw, _ := ioutil.ReadFile("testdata/test.txt") // ignoring error
 	want := string(wantRaw)
-	p := File("testdata/test.txt")
+	p := File("testdata/test.txt\n") // file paths retrieved through Pipe.String() contain a trailing newline
 	gotRaw, err := ioutil.ReadAll(p.Reader)
 	if err != nil {
 		t.Error(err)
@@ -134,7 +134,7 @@ func TestFindFiles(t *testing.T) {
 		Want           string
 	}{
 		{
-			Path:        "testdata/multiple_files",
+			Path:        "testdata/multiple_files\n", // file paths retrieved through Pipe.String() contain a trailing newline
 			ErrExpected: false,
 			Want:        "testdata/multiple_files/1.txt\ntestdata/multiple_files/2.txt\ntestdata/multiple_files/3.tar.zip\n",
 		},
@@ -173,7 +173,7 @@ func TestIfExists(t *testing.T) {
 	if p.Error() == nil {
 		t.Errorf("want error from IfExists on non-existent file, but got nil")
 	}
-	p = IfExists("testdata/empty.txt")
+	p = IfExists("testdata/empty.txt\n") // file paths retrieved through Pipe.String() contain a trailing newline
 	if p.Error() != nil {
 		t.Errorf("want no error from IfExists on existing file, but got %v", p.Error())
 	}
@@ -202,7 +202,7 @@ func TestListFilesNonexistent(t *testing.T) {
 
 func TestListFilesSingle(t *testing.T) {
 	t.Parallel()
-	got, err := ListFiles("testdata/multiple_files/1.txt").String()
+	got, err := ListFiles("testdata/multiple_files/1.txt\n").String() // file paths retrieved through Pipe.String() contain a trailing newline
 	if err != nil {
 		t.Fatal(err)
 	}


sources.go
```diff
@@ -34,6 +34,7 @@ func Exec(s string) *Pipe {
 // status will be set, and if the file does exist, the pipe will have no error
 // status.
 func IfExists(filename string) *Pipe {
+	filename = strings.TrimRight(filename, "\n")
```
bitfield, line 37: Do we need this? If the intention of this PR is only to remove trailing newlines from strings that come from a pipe (for example, `script.Args`), then we don't need this change. I don't think anyone's likely to call `IfExists(pathname)` with a pathname with a trailing newline, are they?

sources.go
```diff
@@ -45,6 +46,7 @@ func IfExists(filename string) *Pipe {
 // starting pipelines. If there is an error opening the file, the pipe's error
 // status will be set.
 func File(name string) *Pipe {
+	name = strings.TrimRight(name, "\n")
```
bitfield, line 49: Same here; the `name` doesn't come from a pipe, so we can leave it as is.

bitfield: It strikes me that it would make more sense for the kind of use case you described to trim trailing newlines in _filter_ methods such as Concat, wouldn't it? But in fact that's not necessary; consider this program:

```go
	script.Args().Concat().Stdout()
```

This works exactly as you'd expect.

sources.go
```diff
@@ -58,6 +60,7 @@ func File(name string) *Pipe {
 // `find -type f`. If the path doesn't exist or can't be read, the pipe's error
 // status will be set.
 func FindFiles(path string) *Pipe {
+	path = strings.TrimRight(path, "\n")
```
bitfield, line 63: Perhaps what we're really saying is that it would be useful for `FindFiles` and friends to be _filters_ as well as sources, isn't it? So we could write, for example:

```go
script.Args().FindFiles().Concat().Stdout()
```

...the effect of which would be to concatenate all files found recursively in the supplied path.

sources.go
```diff
@@ -78,6 +81,7 @@ func FindFiles(path string) *Pipe {
 // supplied path, one per line. The path may be a glob, conforming to
 // filepath.Match syntax.
 func ListFiles(path string) *Pipe {
+	path = strings.TrimRight(path, "\n")
```
bitfield, line 84: Same applies here; if `ListFiles` is not a filter, it doesn't take any pipe input, so it doesn't need any trimming. If it _were_ a filter, I suspect it wouldn't need trimming either; but we can find out.

bitfield (commented):
This is fine as far as it goes, but as I say in the comments, I think maybe if you really want to write this program:

```go
script.Args().ListFiles().Stdout()
```

...then _that_ should be the PR, shouldn't it?

bitfield:
I would also be fine with adding a `Chomp()` method to pipes which is like `String()`, but trims trailing newlines. I think that would be useful too.

christophberger:
Thank you John for your detailed feedback. Obviously you know your lib and its intended way of working so much better than I do (no surprise here). I feel I cannot make any qualified statement about the code and how to best add the requirement ( if it should be added at all), being a completely new user to this package.

From my point of view, the issue boils down to the question about whether functions that take file paths should sanitize their input or whether the user should take care of providing correct input. Bertrand Meyer, for example, decided to put the burden on the caller of a function (and this is why Eiffel has contracts built into the language, to be able to reject wrong input right away rather than trying to fix it). In a language without contracts, I think the decision depends on the user case and library API design. I thus leave it to you to decide, as the one who knows his design best. I would be totally fine to close this PR if you prefer a different route than what the PR suggests.

bitfield:
No worries! I'm a big believer in getting people to be explicit about their desired use cases, to avoid the [XY Problem](https://en.wikipedia.org/wiki/XY_problem). If what you really want is to write the program I quoted:

```go
script.Args().FindFiles().Concat().Stdout()
```

...then I find that elegant and desirable too, and it probably doesn't require any newline trimming. I think `Chomp()` would be useful too, so if it suits you, I'll open two new issues along those lines and anyone who's looking for a bit of Go practice can pick them up!

christophberger:
Sorry for the late reply, I was traveling for a few days.

No, there is definitely no XY problem here. Sure, there are many ways to reach a goal, and if I had been lucky, I would have chosen one that happens to avoid the problem I stumbled over. Still, the original problem exists and probably waits for other people to stumble upon.

Let me explain the original problem in more detail:

This code 

```go
	r, _ := script.Args().String()
	p := script.ListFiles(r)
	p.Stdout()
```

looks valid but prints nothing for any valid path passed as an argument.
Adding error handling only helps if the error output is interpreted correctly:

```go
	r, _ := script.Args().String()
	p := script.ListFiles(r)
	_, err := p.Stdout()
	log.Println(err)
```

```
2020/09/08 13:57:58 stat .
: no such file or directory
```

The newline between the path ("." in this case) and the colon is easily overlooked, or not recognized as part of the path string. Only extra logging with visible string delimiters ("[]") clearly shows where the path string begins and where it ends.

```go
log.Printf("args: [%s]\n", r)
```

```
2020/09/08 13:57:58 args: [.
]
```

So while the solution you propose is certainly the more elegant solution, you cannot avoid that users try the above code and fail. 

The point here is that the output of `Pipe.Stdout()` always ends in a newline, while `FindFiles()` or `ListFiles()` always fail if the path ends in a newline. 

Hence either `FindFiles()` and `ListFiles()` should trim any newlines from their input (thus making them also robust against other sources that provide path strings with newlines), or `Pipe.Stdout()` should remove the final newline from its return value. The latter approach may break existing code, however, so I would suggest that `Find/ListFiles` trim their input.

As I wrote before, using FindFiles() as a filter is certainly an elegant option, and having a `Chomp()` filter may also be welcome to those users who know that they have a string to trim. So I would agree to have new issues for these features but IMHO they don't entirely solve the original problem.

bitfield:
> The point here is that the output of Pipe.Stdout() always ends in a newline, while FindFiles() or ListFiles() always fail if the path ends in a newline.

Yes, I see that. In fact, it's `Args()` which appends newlines to your arguments, because it's turning them into a pipe, which is inherently newline-delimited (otherwise you'd just get all your arguments as a single line).

> Hence either FindFiles() and ListFiles() should trim any newlines from their input

I don't think that's 'hence', exactly. Sure, if you pass a string representing a file path that ends in a newline to _any_ Go function that deals with files, it won't work:

```go
_, err := os.Open("main.go\n")
// open main.go
// : no such file or directory
```

It makes sense for `script` methods to behave the same way as the standard library. So, when `os.Open` trims trailing newlines from its input, I'll do the same!

> you cannot avoid that users try the above code and fail.

I can _never_ avoid people writing incorrect programs, and it would be a mistake to try to design a library that could; it would be very inconvenient for people trying to write _correct_ programs!

> or Pipe.Stdout() should remove the final newline from its return value

Neither String() nor Stdout() add a final newline to their output, so this isn't an issue.

I'll open new issues for the ListFiles/FindFIles filters, and for Chomp(), and close this one, with my thanks to you for reporting it.

christophberger:
Agreed, staying as close as possible to stdlib behavior is never wrong. Also, your pipeline design relies on newlines as separators, and even though the change here would only be to drop the very last newline character from the output of Pipe.Stdout(), I would not want to propose that as a solution as it might break existing client code.

The code I wrote may be incorrect but I wrote it with the knowledge that I had of your library at that time. I was not aware of trailing newlines, so maybe, instead of changing the code, it might be sufficient to add a few words of warnings to the documentation, to prevent `bitfield/script` newbies from trying to stuff the output of a pipeline directly into FindFiles/ListFiles without trimming.

bitfield:
Well, as I say, they'll be able to stuff the output of a pipeline directly into FindFiles/ListFiles when they're implemented as filters! Which is a very good idea, and thank you for leading us to it.

PR #65, opened 2020-10-02
Have FindFiles() take a glob

vigneshashokan:
FindFiles() takes filepath as a glob.

### DIFF
--- a/sources.go
+++ b/sources.go
@@ -59,18 +59,27 @@ func File(name string) *Pipe {
 // status will be set.
 func FindFiles(path string) *Pipe {
 	var fileNames []string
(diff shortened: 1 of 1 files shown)


bitfield (changes requested):
Won't we need some test cases here? If the tests are already passing, which I believe they are, then any non-test code that we write must be by definition untested, mustn't it?

If we're working TDD-style (which I recommend!) then we would start by adding one or more failing test cases which require FindFiles to take a glob. Then we would write the code to make them pass!

bitfield:
I haven't yet seen a use case for this, so closing for now.

PR #70, opened 2020-11-18
Implement Get User Input

latiif:
This PR is my take on #52. It utilizes the same example program and test cases.

The test cases are concurrency safe and this can be checked by running: `go test -race -run TestPrompt`.

This PR only adds prompt functionality to read simple string. One in the future to select from a menu may follow.

In #51, there is a suggestion for a prompt that reads numbers. However I think a sink with the name `Number()` is more appropriate for the functionality similar to what `String()` does.


bitfield:
I wonder if it makes sense to have the prompt message be the contents of the pipe, and to rename this method something like `ReadInput`:

```go
defaultPath := "/usr/local/foo"
path, err := script.Echo("Installation path: ").ReadInput(defaultPath).String()
```

What do you think?

latiif:
It totally makes the prompting functionality more flexible as a `pipe` compared to it as a `source`. I was just following the previous PR's approach. 

The question is should we still have a `source` prompt function that under the hood utilizes the `pipe` `ReadInput(string)`? Or is it considered bloat that way?

Anyhow, I'll keep `Prompt(string, string)` as a `source` for the time being and start working on `ReadInput(string)` the `pipe`.

examples/prompt/main.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+func main() {
+	script.Echo("Installation of Fake Product\n").Stdout()
+	curDir, err := script.Exec("pwd").String()
+	if err != nil {
+		panic(err)
+	}
+
+	script.Prompt("Choose install location", strings.TrimSpace(curDir)).Stdout()
```
bitfield, line 16: Suppose you use the install location as an argument to a fake `install` command, for example?

bitfield (commented):
This looks pretty sound to me; I wouldn't worry about retaining `Prompt`, since you've effectively made it redundant, haven't you?

examples/prompt/main.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import (
+	"strings"
+
+	"github.com/bitfield/script"
+)
+
+func Install(path string) {
+	script.Echo("📂 Installing at " + path + "\n").Stdout()
```
latiif, line 10: We really need `Echoln` and `Echof`  :)

README.md
```diff
@@ -881,6 +890,7 @@ Since `script` is designed to help you write system administration programs, a f
 * [head](examples/head/main.go)
 * [least_freq](examples/least_freq/main.go)
 * [ls](examples/ls/main.go)
+* [prompt](examples/prompt/main.go)
```
bitfield, line 893: The shell builtin is called `read`, isn't it—perhaps we should use that name.

examples/prompt/main.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+func main() {
+	script.Echo("Installation of Fake Product\n").Stdout()
+	curDir, err := script.Exec("pwd").String()
+	if err != nil {
+		panic(err)
+	}
+
+	path, _ := script.Echo("Choose install location").ReadInput(strings.TrimSpace(curDir)).String()
```
bitfield, line 21: We can use `ExecForEach` here, can't we?

README.md
bitfield: (Shall we add this to the table of Unix equivalents as well, under `read`?)

filters.go
```diff
@@ -311,6 +311,38 @@ func (p *Pipe) MatchRegexp(re *regexp.Regexp) *Pipe {
 	})
 }
 
+// ReadInput reads displays the pipe content as a message as a prompt
```
bitfield, line 314: One question that occurs with this is how to test it! Hard-wiring `os.Stdin` is not great for several reasons, but this is a larger issue with `script` in general; I'd like the pipe's standard in and standard out to be reassignable. When we have that, this issue will go away.

filters.go
```diff
@@ -311,6 +311,38 @@ func (p *Pipe) MatchRegexp(re *regexp.Regexp) *Pipe {
...
+// and reads the content from standard input and returns the result
+// as a pipe. In case the input was empty it defaults to its `defaultValue`
+func (p *Pipe) ReadInput(defaultValue string) *Pipe {
+	const emptyString = ""
```
bitfield, line 318: There's really no value in defining a constant for empty string, is there? We can just use an empty literal instead.

filters.go
```diff
@@ -311,6 +311,38 @@ func (p *Pipe) MatchRegexp(re *regexp.Regexp) *Pipe {
...
+	const withDefault = "%s [%s]: "
```
bitfield, line 319: I don't think these constants are useful either; in fact, they make the code obscure, because when you're looking at line 331, for example, you can't see what `withDefault` actually is. You have to jump elsewhere in the source to find it and then jump back. That's a speed bump for the reader.

filters.go
```diff
@@ -311,6 +311,38 @@ func (p *Pipe) MatchRegexp(re *regexp.Regexp) *Pipe {
...
+	const withoutDefault = "%s: "
+	if p == nil || p.Error() != nil {
+		return p
+	}
+	defer p.Close()
```
bitfield, line 324: We don't want to close the pipe, do we?

filters.go
```diff
@@ -311,6 +311,38 @@ func (p *Pipe) MatchRegexp(re *regexp.Regexp) *Pipe {
...
+
+	prompt, _ := p.String() // if pipe is empty, prompt message is empty
```
bitfield, line 326: We need to check error here.

filters.go
```diff
@@ -311,6 +311,38 @@ func (p *Pipe) MatchRegexp(re *regexp.Regexp) *Pipe {
...
+
+	if defaultValue == emptyString {
```
bitfield, line 328: Can't we do a little better here? This seems rather repetitive.

filters.go
```diff
@@ -311,6 +311,38 @@ func (p *Pipe) MatchRegexp(re *regexp.Regexp) *Pipe {
...
+		Echo(fmt.Sprintf(withoutDefault, prompt)).Stdout()
+	} else {
+		Echo(fmt.Sprintf(withDefault, prompt, defaultValue)).Stdout()
+	}
+
+	scanner := bufio.NewScanner(Stdin().Reader)
```
bitfield, line 334: Did you think about something like `fmt.Scanln` instead here?

filters_test.go
```diff
@@ -346,6 +348,68 @@ func TestMatch(t *testing.T) {
 	}
 }
 
+func TestReadInput(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		Description string
```
bitfield, line 354: This doesn't match the look and feel of test cases in other tests.

filters_test.go
```diff
@@ -346,6 +348,68 @@ func TestMatch(t *testing.T) {
...
+			Input:       "/user/test @@@ testing one two three",
+			Default:     "default",
+			Want:        "/user/test @@@ testing one two three",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.Description, func(t *testing.T) {
+			// Prepare Stdin before calling prompt.
+			tmp, err := ioutil.TempFile("", tc.Description)
```
bitfield, line 383: A `bytes.Buffer` would probably be more convenient here, wouldn't it?

filters_test.go
```diff
@@ -346,6 +348,68 @@ func TestMatch(t *testing.T) {
...
+			if _, err := tmp.WriteString(tc.Input); err != nil {
+				t.Error(err)
+			}
+
+			if _, err := tmp.Seek(0, 0); err != nil {
+				t.Error(err)
+			}
+
+			// Restore original Stdin.
+			oldStdin := os.Stdin
```
bitfield, line 397: This definitely isn't concurrency-safe! Have a look at `TestStdin`, for example, for another approach. As I say, I'd like to solve the whole stdin/stdout issue globally for pipes, but we'll get there.

filters_test.go
```diff
@@ -346,6 +348,68 @@ func TestMatch(t *testing.T) {
...
 }
 
+func TestReadInput(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		Description string
+		Input       string
+		Default     string
+		Want        string
+		ShouldFail  bool
```
bitfield, line 358: How about `wantErr`, for example, to match other tests?

bitfield (commented):
I think we've got a working approach now, haven't we? Nice job!

README.md
latiif: [_Nja_](https://en.wiktionary.org/wiki/nja#Swedish), while `read` is similar in function and behavior, it's not really exactly what `ReadInput` does here. For example `read DIR_NAME` doesn't directly pipe user input to the next command. It doesn't really have a built-in way for default input either.

Having said that, I agree that adding `read` to the list might help new comers use `ReadInput` so I suggest writing `read*` and explain the differences somehow.

examples/prompt/main.go
latiif: I don't really follow, what would `ExecForEach` do here? `ReadInput` reads only one line (reads off `stdin` and stops at newline).

Or do you mean instead of `String()`? Something like:

```diff
diff --git a/examples/read/main.go b/examples/read/main.go
index 55ab3a9..10c31cd 100644
--- a/examples/read/main.go
+++ b/examples/read/main.go
@@ -18,6 +18,6 @@ func main() {
 		panic(err)
 	}
 
-	path, _ := script.Echo("Choose install location").ReadInput(strings.TrimSpace(curDir)).String()
+	path, _ := script.Echo("Choose install location").ReadInput(strings.TrimSpace(curDir)).ExecForEach("echo {{.}}").String()
 	Install(path)
 }
```

I might be misinterpreting your comment, but adding `ExecForEach` just for the sake of using it, might throw off people trying to learn more about `ReadInput`.

filters.go
latiif: That's a great point. I personally didn't like it either to hard wire `ReadInput` to `os.Stdin` but as you pointed out, solving it seemed a bit out of scope.

My suggestion is creating two `ReadInput` versions, (since Go doesn't allow function overloading or default parameters).

One package-scoped and private 
```go
func (p *Pipe) readInput(defaultValue string, stdin io.Reader) *Pipe
```
And the public one:
```go
func (p *Pipe) ReadInput(defaultValue string) *Pipe {
    return p.readInput(defaultValue, os.Stdin())
}
```

And in the tests, we will have to convince ourselves that testing `readInput` is good enough. And that this compromise will save us the trouble of stubbing `os.Stdin()` in a safe way.

filters.go
latiif: Of course, I was just thinking that reading the code would flow a bit better.

filters.go
latiif: Fair point.

However, I think that given the relative small size of the function body; I'd say defining constants is good actually for readability — the maintainer doesn't really need to know what the actual strings are. Plus if we have to reuse the constants they're easy-to-find and to change.

Defining `withDefault` and `withoutDefault` up-first; gives a good contrast as to what they semantically represent.

If you still think it's better to use plain literals; I'm happy to do the change.

filters.go
latiif: I considered and tried it out, but ultimately decided against it. 

First, `fmt.Scanln` and the entire `fmt.Scan*` family seem to scan space-separated tokens. Plus, `bufio.NewScanner(io.Reader)` is easier to stub and be abstracted.

filters_test.go
latiif: I agree. But I chose to follow the canon example of parallel table driven testing as demonstrated in [the golang wiki](https://github.com/golang/go/wiki/TableDrivenTests#parallel-testing).

I personally think it's better if the other tests followed _this_ format.

filters_test.go
latiif: For which function exactly? Both functions `t.Run` and `ioutil.Tempfile` take `string`.

```go
func (t *T) Run(name string, f func(t *T)) bool 
func TempFile(dir, pattern string) (f *os.File, err error)
```

Please, correct me if I misunderstood.

filters_test.go
latiif: Agreed.

filters_test.go
latiif: I am all for solving the problem in a different way. However, If we decide against doing the solution discussed earlier (the one with two versions). I'm all for stubbing `os.Stdin` and using a `mutex` to ensure concurrency-safety. 

The suggested solution in `TestStdin` while apparently concurrency-safe seemed a bit impure (dependency on external shell commands if I understood correctly). While the task is admittedly complex, I am really all for solving it differently. 

I ran `TestReadInput` with Golang's race detector with the command below, and got no errors. I am no expert in concurrency but I found it strange that the race detector didn't detect the error, or is it just that it wasn't made to discover that exact type.
```bash
go test -count=1 -race -v -run TestReadInput 
```

README.md
bitfield: Well, that table is really to help people who are in the position of "I have some existing shell script, or something that I know how to do in shell, and I want to do it with `script`—what should I look for?" For example, if you want to read user input, you might reach for the `read` builtin, and we want you to be guided towards `ReadInput`, even though it's not an exact analogue.

examples/prompt/main.go
bitfield: Well, if we just wanted to read a string as input, we could use `fmt.Scanf`, couldn't we? The value of having it as a pipe method is precisely that we can then _use_ that input in the next stage of the pipe!

In the installer example, we might read the installation path and then pass it to the fictitious `install` command:

```go
script.Echo("install location: ").ReadInput("").ExecForEach("/usr/bin/install {{.}}").Stdout()
```

filters.go
bitfield: I'm using `TestMain` to do this for some tests (see `TestStdin`, for example).

filters.go
bitfield: Well, if we define the constants right next to where they're used, it's just as easy to find and change the _literals_ instead, isn't it?

The only real value in defining a constant is so that it can be a long way away from the place where it's used. But we _don't_ want that here.

filters_test.go
bitfield: Sorry, I should have been more specific: I meant that the field names are not exported, for example (and it's the same in the Go wiki you referenced).

What format were you referring to, then?

filters_test.go
bitfield: Well, the point is that we just need some `io.Reader`, isn't it? We could use a disk file for this purpose, but it's tremendously slow compared to anything in memory (like a `bytes.Buffer`), and it also requires us to have a writeable filesystem, which isn't necessarily always the case (think about running tests in a container, for example).

filters_test.go
bitfield: The race detector can only detect races that actually _occur_ while running the tests, not potential races. I think if you create _another_ parallel test that tries to write to `os.Stdin`, for example, you might trigger it.

Okay, it sounds like this is where we are: we don't want to do stdin-related testing in a crap way, and we don't want to put a bunch of engineering effort into doing it a better way that won't be needed once we've solved the overall issue of pipe-level stdin/stdout. So we should probably solve that _first_, and then come back to this PR and use it.

Are you up for having a go?

README.md
latiif: Done, please resolve.

examples/prompt/main.go
latiif: Agreed. I however used `echo Installing at {{.}}` because it has an output. 

When I ran with `/usr/bin/install {{.}}` the example program didn't output anything. I figured having an `echo` makes the example more lively.

filters.go
latiif: IMHO, this discussion seems to not have an objective answer. I'm more than glad to follow the project conventions and use the literals directly.

filters.go
latiif: I did my best effort following the example of `TestStdin` and keeping the tests thread-safe

filters_test.go
latiif: Sorry, I got things mixed up. Fixed, samll letters are used as per the guide.

filters_test.go
latiif: Great points there. I guess the latest commit addresses these issues.

bitfield:
Notwithstanding the excellent work done on this, I'm closing the PR with some reluctance, because I just can't seem to fit it into the general scheme of the `script` library. It's essentially about _pipelines_, and I haven't seen a use case where 'read from terminal' can reasonably be part of a pipeline. (The shell builtin `read` is never used this way, either.)

PR #72, opened 2021-01-11
DRAFT: Initial PoC of a HTTP sink and source

hikhvar:
PoC to explore the design proposed in #73 . This PR does not have test or docs.


bitfield:
@hikhvar I've been doing some (much-belated, sorry about that) thinking about this. HTTP support is a really valuable feature to add to `script`, but as you know, I have a relentless obsession with simplicity and it's not clear how to do this in an obsessively simple way. I'd like to mull it over a little more, so I'll close this in favour of #85.

However, I _do_ like the `JQ` idea and have come round to thinking that it should be a native `Pipe` method. Would you mind if I open a new PR just for that, using your `JQ` implementation as a basis?

Issue #74, opened 2021-01-11
Add a jq filter


hikhvar:
Modern shell scripts interact often with a REST API or a command line program producing json (kubectl for example). Slicing and dicing the data returned by those APIs is often done with https://stedolan.github.io/jq/. 

https://github.com/itchyny/gojq provides a go native implementation of jq.

I propose the following filter:
```
func (p *Pipe) JQ(query string) *Pipe
```

An example script:
```
script.Echo('{"foo": 128}'').JQ(".foo").Stdout()
```
should return "128" on stdout. 

Or to show the IP addresses of the local interface `lo` a shell user will write:
```
ip -j a show  | jq '.[] | select(.ifname=="lo") | .addr_info[].local'
```
With an jq filter in `script` we can write:
```
 script.Exec("ip -j a show").JQ('.[] | select(.ifname=="lo") | .addr_info[].local').Stdout)
 ```
 
 The same slicing and dicing of data can be done in pure go code, but the jq language provides a well documented and known DSL for this problem.

bitfield:
This is a very neat idea! I can imagine doing something similar with XPath, for example. I'm not sure if `JQ` is the right name for this—how about `JSONQuery`, perhaps?

hikhvar:
I'm fine with renaming it to JSONQuery. However there is a JSONQuery DSL available in the NPM ecosystem: https://www.npmjs.com/package/json-query 

Also googling JSONQuery lead to a lot of json query languages not equal to the language implemented by `jq`.

Providing filters for XPath oder yq (yaml equivalent of jq) adds value too. Do we need all those together in the same design?

bitfield:
Yes, the only problem with `JQ` is that it doesn't tell you what it does, if you're not already familiar with `jq`, for example. This is why we have `Column`, not `Cut`, and `First`, not `Head`. But, like you, I can't think of a better name. `JQ` is fine for now.

Yes, XPath would be nice, but I'll open a separate issue for that.

Can we write a non-trivial `script` program which demonstrates the value of a `JQ` method over, say, just piping stuff through `jq` using an `Exec`?

hikhvar:
[...]

PR #75, opened 2021-01-21
Add ExternalFilter filter to enable the user to write custom filter

hikhvar:
This enables the user to write own filters and use them in the fluent interface of the Pipe without forking the core library.

Implement the ExternalFilter function as discussed in #74

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ examples/visitors/visitors
 examples/*/go.sum
 .vscode/settings.json
 examples/ls/ls
+.idea
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ John is also the author of the popular [For the Love of Go](https://bitfieldcons
 	- [EachLine](#eachline)
 	- [Exec](#exec-1)
 	- [ExecForEach](#execforeach)
+	- [ExternalFilter](#externalfilter)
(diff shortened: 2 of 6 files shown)


hikhvar:
@bitfield can you please take a look at this PR? Anything missing?

bitfield:
I haven't forgotten about this @hikhvar: it's significant enough that I need to make time and space to look at it very carefully and think about it hard. I'd like to try it out myself with some other use case and see how it feels in action!

lalloni:
@hikhvar this looks pretty good! I'd love to see it merged.

One thing I would question is calling it "external" given that external sounds more like calling out to something external to the program. So I'd suggest to change the name to something like `FuncFilter` or just `Func` because it would follow more closely the rest of the API style.

hikhvar:
@lalloni I like your suggestion. I renamed the function to FilterFunc.
@bitfield can you please have a second look?

bitfield:
This dream is now realised, though I've gone a slightly different way with it, which I think is a bit more general: the filter func takes a reader and a writer, and returns error. Since it was your idea, I've added you as a co-author on that commit—I hope that's okay with you.

hikhvar:
I'm very ok with this! Really like the more general interface.

Issue #61, opened 2020-08-28
Execute with an argument list


NightMachinary:
What is Exec doing under the hood? Is it invoking a shell, or is it parsing the string itself? And how can we use an explicit argument list and not a string?

Generally, I think a major shortcoming of the readme is that it does not clarify the security considerations of using this library. Shell is very insecure, and one of the major reasons not to use it is to have better security.

Update: I now see this is kind of a duplicate: https://github.com/bitfield/script/issues/32

bitfield:
Hey @NightMachinary, thanks for the question!

> Is it invoking a shell, or is it parsing the string itself?

It uses Go's standard library `exec.Command` to run the external command directly (not via a shell). Though you can, of course, exec a shell if you like and give it a command to run.

https://github.com/bitfield/script/blob/master/filters.go#L120

> how can we use an explicit argument list and not a string?

Can you clarify the question slightly for me? Are you asking if there's a way to execute a command with a _slice_ of arguments, instead of a space-separated string? If so, something like this might work:

```go
args := []string{"hello", "world"}
p := script.Exec("echo " + strings.Join(args, " "))
```

It's usually more convenient to _start_ with a string, which is why Exec takes a string instead of a slice. If you have a use case where it's more convenient to pass a slice, I'd like to hear about it. We could provide that as an alternative.

> Generally, I think a major shortcoming of the readme is that it does not clarify the security considerations of using this library.

Fair enough. What do you think needs to be mentioned in the README about this? I'm happy to accept PRs or to make the updates myself.

NightMachinary:
[**@bitfield**](https://github.com/bitfield)
> Can you clarify the question slightly for me? Are you asking if there's a way to execute a command with a _slice_ of arguments, instead of a space-separated string? If so, something like this might work:
> 
> ```go
> args := []string{"hello", "world"}
> p := script.Exec("echo " + strings.Join(args, " "))
> ```
> 
> It's usually more convenient to _start_ with a string, which is why Exec takes a string instead of a slice. If you have a use case where it's more convenient to pass a slice, I'd like to hear about it. We could provide that as an alternative.

Yes, having a slice is sometimes more convenient; Generally, whenever the command is going to use some dynamic variable, there would be a danger of that variable containing spaces. For example, consider a fictional program `get-movie-rating` which takes exactly one argument, the name of a movie. `script.Exec("get-movie-rating " + movie_name)` would fail if the name contains whitespace.

[...]

PR #76, opened 2021-02-25
Add ExecArgs source and filter

lalloni:
Would help to solve #61.

### DIFF
--- a/filters.go
+++ b/filters.go
@@ -124,12 +124,22 @@ func (p *Pipe) Exec(cmdLine string) *Pipe {
 	if p == nil || p.Error() != nil {
 		return p
 	}
-	q := NewPipe()
 	args, ok := shell.Split(cmdLine) // strings.Fields doesn't handle quotes
 	if !ok {
 		return p.WithError(fmt.Errorf("unbalanced quotes or backslashes in [%s]", cmdLine))
 	}
-	cmd := exec.Command(args[0], args[1:]...)
+	return p.ExecArgs(args[0], args[1:]...)
+}
+
+// ExecArgs runs an external command and returns a pipe containing the output. If
+// the command had a non-zero exit status, the pipe's error status will also be
+// set to the string "exit status X", where X is the integer exit status.
+func (p *Pipe) ExecArgs(command string, args ...string) *Pipe {
+	if p == nil || p.Error() != nil {
+		return p
+	}
+	q := NewPipe()
+	cmd := exec.Command(command, args...)
 	cmd.Stdin = p.Reader
 	output, err := cmd.CombinedOutput()
 	if err != nil {
--- a/filters_test.go
+++ b/filters_test.go
@@ -149,6 +149,32 @@ func TestExecFilter(t *testing.T) {
 	}
 }
 
+func TestExecArgsFilter(t *testing.T) {
+	t.Parallel()
+	want := "a message with spaces"
+	p := File("testdata/hello.txt")
+	q := p.ExecArgs("echo", "-n", "a message with spaces")
+	got, err := q.String()
+	if err != nil {
+		t.Error(err)
+	}
+	if got != want {
+		t.Errorf("want %q, got %q", want, got)
+	}
+	// This should fail because p is now closed.
+	_, err = p.String()
+	if err == nil {
+		t.Error("input not closed after reading")
+	}
+	p = Echo("hello world")
+	p.SetError(errors.New("oh no"))
+	// This should be a no-op because the pipe has error status.
+	out, _ := p.Exec("cat").String()
+	if out != "" {
+		t.Error("want exec on erroneous pipe to be a no-op, but it wasn't")
+	}
+}
+
 func TestExecForEach(t *testing.T) {
 	t.Parallel()
 	tcs := []struct {
--- a/sources.go
+++ b/sources.go
@@ -29,6 +29,13 @@ func Exec(s string) *Pipe {
 	return NewPipe().Exec(s)
 }
 
+// ExecArgs runs an external command and returns a pipe containing the output. If
+// the command had a non-zero exit status, the pipe's error status will also be
+// set to the string "exit status X", where X is the integer exit status.
+func ExecArgs(cmd string, args ...string) *Pipe {
+	return NewPipe().ExecArgs(cmd, args...)
+}
+
 // IfExists tests whether the specified file exists, and returns a pipe whose
 // error status reflects the result. If the file doesn't exist, the pipe's error
 // status will be set, and if the file does exist, the pipe will have no error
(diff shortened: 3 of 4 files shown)


lalloni:
Hey @bitfield, any chance of this being considered?

bitfield:
Not at the moment, @lalloni: I mentioned in a comment on your issue that I haven't yet seen a problem that this solves. If you have one, of course I'll be pleased to see the program that demonstrates it.

lalloni:
When building long command lines with more than a few computed arguments to pass to it, the requirement of using shell style quoted syntax makes the library harder to use and introduces a opportunities to do it wrong (like forgetting to quote arguments that could contain spaces).

For example, let's take the execution of the go tool to build a binary...

Basic shell execution would be like:

```sh
go build -ldflags '-X main.version=1.2.3' -o 'build/some-program' 'cmd/some-program/main.go'
```

If we build that line from a go program that executes the go tool to build all the binaries in a given project, it would naturally be something as simple as:

```go
out, err := foo.Exec("go", "build", "-ldflags", "-X main.version=1.2.3", "-o", "build/some-program", "cmd/some-program/main.go").String()
// do something with err/out, etc...
```

Note that the strings passed above do not need any further processing before passing them to the command execution function used to run the actual command (either `os.StartProcess()` or `os/exec.Command()`), and they are also in the right granularity so avoiding any quoting to demark arguments boundaries.

Of course since our build tool would need to potentially build many binaries we need to introduce some variables to loop over the `cmd/` subdirectories and to compute where to write each final binary, as well as for the version to use:

```go
out, err := foo.Exec("go", "build", "-ldflags", "-X main.version="+version, "-o", filepath.Join(buildDir, programName), filepath.Join(cmdDir, "main.go")).String()
// do something with err/out, etc...
```

Which is the minimum we need to write without introducing complexities related to shell syntax and quoting (as you can see there is no need at all to parse these strings looking for quoting, `foo.Exec` only needs to pass the `[]string` it receives to `os/exec.Command()` or even to `os.StartProcess()` without any kind of massage.

In contrast, using `bitfield/script.Exec()` requires the code to compute a single string with proper quoting and compatible syntax so it can at its turn parse it and split it back to pass it to the underlying command execution function, for example:

```go
out, err := script.Exec("go build -ldflags '-X main.version="+version+"' -o '"+ filepath.Join(BuildDir, binary) +"' '" + filepath.Join(".", binary, "main.go")).String()
// do something with err/out, etc...
```

Which is way harder to write, to not make any mistakes on the quoting, and it is also less efficient since we are "inventing" a string just to satisfy the `script.Exec()` function that will need to parse it using some (arbitrary) syntax to be able to split it again back to get the `[]string` that we already started with in the first place!

So what my patch above does is simply adding an `Exec` function variant that can receive the strings that compose the command as a natural `[]string` and it simply passes it verbatim to the underlying command execution function. This saves the coder needing to build a quoted string which is awkward and error-prone and it also saves the library from parsing it just to split it back to the original slice form.

lalloni:
So, @bitfield, the problem it solves is about needing to build a synthetic string using some shell-like syntax to demark the target command arguments when we originally have the simple `[]string` that we need to pass to the underlying execution function.

bitfield:
How about something like this:

```go
objName := filepath.Join(BuildDir, binary)
sourcePath := filepath.Join(".", binary, "main.go")
buildCmd := fmt.Sprintf("go build -ldflags '-X main.version=%s' -o %q %q", version, objName, sourcePath)
out, err := script.Exec(buildCmd).String()
```

Does that solve your problem?

lalloni:
Sure @bitfield, that is an alternative, as well as other possible additions we could make in the form of other helper functions to reduce the "boilerplate" code needed to build the longer command string using shell-like syntax... but that is besides the point.

The point is that the simplest possible thing is to simply receive the `[]string` and pass it as is to the underlying command execution function... we are adding unneeded complexity (on both client code and internal code) by demanding the use of the shell-like syntax... it's like we are in love with the idea of parsing/supporting such pattern as the only way of executing commands, and unfortunately is not the simplest one, and it not only adds complexity and inefficiency but also error-making opportunities.

bitfield:
Sorry, but I'm not convinced by this. The `script` method that runs a command takes a single string, and if it's easier for you to build it up as a `[]string` for some reason, you can always use `strings.Join` to concatenate it. Even better, as I suggested, is to use `fmt.Sprintf` to interpolate data into the required string. (`script` also has `ExecForEach` for running the same command on lots of computed values.)

> we are adding unneeded complexity

We are refraining from adding unneeded complexity. We don't need two different ways to run commands; that is confusing for beginners, and superfluous for experienced programmers. Thanks for the suggestion, but there's a high bar to increasing the size of the `script` API, and while this is a good idea, it doesn't quite show enough benefit for the cost.

kudenz:
Thank to tips

Pada tanggal Rab, 21 Apr 2021 19.58, Pablo Lalloni ***@***.***>
menulis:

lalloni:
> Sorry, but I'm not convinced by this. The `script` method that runs a command takes a single string, and if it's easier for you to build it up as a `[]string` for some reason, you can always use `strings.Join` to concatenate it. Even better, as I suggested, is to use `fmt.Sprintf` to interpolate data into the required string. (`script` also has `ExecForEach` for running the same command on lots of computed values.)

`strings.Join` helps little because we need to quote, `fmt.Sprintf` isn't enough when we don't know the number of elements (not to mention we need to write "code" with a "format" strings like `"%q %q %q %q"` which is just plain... a bother and error prone)... in both cases we would scramble to add a helper functions to save all of that.

> We are refraining from adding unneeded complexity. We don't need two different ways to run commands; that is confusing for beginners, and superfluous for experienced programmers. Thanks for the suggestion, but there's a high bar to increasing the size of the `script` API, and while this is a good idea, it doesn't quite show enough benefit for the cost.

Unfortunately, the way that was chosen to run commands is not the simplest possible one because instead of focusing in running commands it also attempts to parse some command arbitrary syntax and that is what creates this discussion. That's the added complexity. 

The simplest possible function to run commands would just receive a `[]string` and would not try to invent/parse a syntax or require client code to deal with it.

Quoting/Unquoting, parsing, etc. is a shell's business... not a command execution function's, and much less it is something I would like to deal with as a client code writer.

If I wanted to execute something as simple as `cut -d "` using this smarty pants unquoting execution function I'd be forced  to write something like:

```go
script.Exec(fmt.Sprintf("cut -d %q", "\""))
```

Which is totally unnecessary, when we could simply say:

```go
script.Exec("cut", "-d", "\"")
```

Just as we would when using the standard library with:

```go
exec.Command("cut", "-d", "\"")
```

bitfield:
I think this will do what you want:

```go
script.Exec("cut -d \"")
```

Issue #34, opened 2019-10-23
Stream pipes


posener:
I really like the idea of this library and it might makes a lot of scripting tasks easy!

The current implementation read all output of each command into memory, and then passes it to the next command and so forth.

For example:
* The usage of [`CombinedOutput`](https://github.com/bitfield/script/blob/master/filters.go#L119) (Also not sure why it was chosen to use the combined output and not stdout?)
* The usage of [string builder in the ForEachLine](https://github.com/bitfield/script/blob/master/filters.go#L97).

It would be nice to avoid it if it is not necessery. Let each reader read as much as needed from previous stage and output whatever necessary to the next stage. This is also how pipes in bash works.

posener:
I've re-written some of the functionality of this library with streaming enabled to explain better what I was thinking: https://github.com/posener/script.

bitfield:
Sounds great! How about putting in a PR to this library to add the streaming functionality?

posener:
Thanks! This is not new functionality but rather replacing the whole underlying mechanism...

bitfield:
Fine! How about putting in a PR to replace the whole underlying mechanism?

bitfield:
I think this is worth doing.

kepkin:
any news on that?

bitfield:
@kepkin it needs a little thought—are you having an issue with memory use?

kepkin:
Well.. I see a totally different approach.

Bash is actually starts all programs in pipeline asynchronously. We can repeat that in go and get following bonuses:
 - low memory usage cause data is piped
 - faster execution as you don't need to accumulate data before passing it further
 - you can make fancy stuff like watch (see GIF in my repo)

Please check out:
https://github.com/kepkin/async-script/tree/master

posener:
@kepkin check out https://github.com/posener/script that was inspired from this project, has similar functionality but was designed for streaming the data.

bitfield:
It's interesting to see these efforts, but I do have the strong feeling that they're solving a problem which does not in fact exist (or, at any rate, they haven't demonstrated that a problem exists). In general, I prefer simple, natural, obvious design, even if that's inherently inefficient. Because the cost of complexity affects every single feature and every contributor to the project, now and in the future, I set a very high bar for making such changes.

For devops and sysadmin tasks, which I imagine is where the `script` library is most useful, performance is very rarely an issue. (I'd be delighted to see a counterexample, though.) 

[...]

Issue #78, opened 2021-04-03
Real-time output to stdout in a long-running command


besi1z:
e.g. for command like

```go
script.Exec("ping -c 100 www.google.com").Stdout()
```

I expect the result shows in stdout line by line just like real Bash script, instead of showing after the run.

Do you have tricks to get this done? or any plan/feature to implement this?

bitfield:
That's a good suggestion! How do you envisage this looking from the code point of view? What `script` pipeline would you write to get real-time output?

marco-m:
One possibility would be:
```go
script.Exec("ping -c 100 www.google.com").Stream()
```
and the implementation could just use StdoutPipe() and StderrPipe() of package os.exec, something along the lines of (didn't check the source of script.Exec, this code doesn't follow script's design with Pipe, it is just to give an idea):
```go
        cmd := exec.Command(...)

	stdout, err := cmd.StdoutPipe()
	if err != nil {
		return err
	}
	stderr, err := cmd.StderrPipe()
	if err != nil {
		return err
	}

	if err := cmd.Start(); err != nil {
		return err
	}

	scanOut := bufio.NewScanner(stdout)
	for scanOut.Scan() {
		line := scanOut.Text()
		if line != "" {
			fmt.Println(line)
		}
	}
	scanErr := bufio.NewScanner(stderr)
	for scanErr.Scan() {
		line := scanErr.Text()
		if line != "" {
			fmt.Println(line)
		}
	}

        // maybe here we need something smarter and collect the errors so that we always call cmd.Wait()
	if err := scanOut.Err(); err != nil {
		return err
	}
	if err := scanErr.Err(); err != nil {
		return err
	}

	if err := cmd.Wait(); err != nil {
		return err
	}
```

bitfield:
That sounds promising! Do you want to put a PR together along these lines so we can try it out?

bitfield:
Now implemented in v0.19.0.

madAndroid:
Has this actually been implemented? I see that the PR was closed and not merged, and I don't see anything in the docs around streaming of output.. is this still something that would be implemented in time? It would be really helpful to have a Go variant of something like Python's `subprocess`

bitfield:
Yes, indeed. Here's what the [README](https://github.com/bitfield/script#filters) says:

> Note that filters run concurrently, rather than producing nothing until each stage has fully read its input. This is convenient for executing long-running comands, for example. If you do need to wait for the pipeline to complete, call `Wait`.

[...]

Issue #59, opened 2020-07-29
EachLine only executed after Exec exits?


xxxserxxx:
It appears that `EachLine()` is only called after a pipe is closed.  This prevents using `script` to write daemon scripts.

If there's a way to perform the equivalent of:

```
dbus-monitor --system "interface=org.freedesktop.DBus.Properties" | \
    while read -r line; do
        printf "DBus event: %s\n" "$line"
    done
```

would it be possible to get an example?  If not, can we change the behavior of `EachLine()` to process each line as it is received, or add an `EachLineForever()` function?

bitfield:
Thanks @xxxserxxx! Could you supply a program that demonstrates the behaviour you're talking about?

xxxserxxx:
I have a bunch of little bash scripts that are long-running processes in the background.  One of them watches for DBus events and executes a command when it catches a match:

```
#!/bin/bash

DE=$(/usr/bin/hcitool con | grep handle | awk '{print $3}' | tr ':' '_')
INTERFACE=org.freedesktop.DBus.Properties
MEMBER=PropertiesChanged
PTH=/org/bluez/hci0/dev_${DE}

/usr/bin/dbus-monitor --system "path='$PTH',interface='$INTERFACE',member='$MEMBER'" |
        while read -r line; do
                echo $line | /usr/bin/grep -q "boolean true"
                if [[ $? -eq 0 ]]; then
                        HA=$(/usr/bin/hcitool con | grep handle | awk '{print $5}')
                        /usr/bin/hcitool lecup --handle ${HA} --min 6 --max 7 --latency 0
                fi
        done
```

How do you represent any script that does not exit, but continually processes output from a command?  Another example would be a script that watches on the output from `dmesg` and performs some action; or one that `tail -F ~/.xsession-error` and performs some action whenever a line matches.

bitfield:
Ah, I see what you're getting at now! This is a good point. As you say, each stage of a pipeline usually reads its input to completion, rather than reading a line at a time. This relates to #34, I think; if we had this kind of 'lazy reading' functionality, you could implement Unix `tail -f`, for example.

I can see what's needed, but I freely admit that right now I don't know _how_ to do it. This needs some thought. I'd welcome suggestions!

xxxserxxx:
@bitfield 

```
package main

import (
	"bufio"
	"fmt"
	"os/exec"
	"time"
)

func main() {
	cmd := exec.Command("cat")
	inpipe, err := cmd.StdinPipe()
[...]

PR #88, opened 2021-11-26
pipeline streaming

gty929:
This PR is an attempt to tackle issue #34. I used io.Pipe() in the `EachLine()` to implement streaming, which is simple and efficient.

As an example, the following program outputs 'toc' five times, with one second time interval:
```go
script.Slice(make([]string, 5)).ExecForEach("bash -c 'echo tic; sleep 1'").Replace("tic", "toc").Stdout()
```

Unluckily, the program is not backward compatible. A user must append a Sink method at the end of the pipeline if s/he wants to wait until the program finishes. For this reason, I added a Wait() method, and slightly modified the test for `ExecForEach()`. (The code passed all other tests.) Another problem is that the `strings.Builder` in `EachLine()` cannot be reset, since the stream is append only.
Currently, Error() and ExitStatus() return the intermediate status of a running pipe, since they are not Sink methods.

Probably, one solution is to add a bool flag in the `Pipe` struct to enable/disable streaming. (Disabled by default & use a Stream() source method to enable it). Should I go ahead?

### DIFF
--- a/README.md
+++ b/README.md
@@ -859,6 +859,15 @@ fmt.Println(err)
 // Output: read test.txt: file already closed
 ```
 
+## Wait
+
+`Wait()` waits until the pipe finishes, consumes it, and returns the error:
+
+```go
+err := script.Slice(make([]string, 5)).ExecForEach("sleep 1").Wait()
+// should sleep for 5 seconds
+```
+
 ## WriteFile
 
 `WriteFile()` writes the contents of the pipe to a named file. It returns the number of bytes written, or an error:
--- a/examples/stream/main.go
+++ b/examples/stream/main.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/bitfield/script"
+)
+
+// This program runs tests on the script library 50 times and prints out the progress nicely.
+
+func main() {
+	round, step := 10, 5
+	progressInfo := make([]string, round)
+	for i := 0; i < round; i++ {
+		progressInfo[i] = fmt.Sprintf("------ Done %v / %v ------", (i+1)*step, round*step)
+	}
+	cmd := fmt.Sprintf("bash -c 'go test -count %v github.com/bitfield/script; echo {{.}}'", step)
+	// with Stream(), the program can print to stdout in real time
+	script.Slice(progressInfo).Stream().ExecForEach(cmd).Stdout()
+}
--- a/filters.go
+++ b/filters.go
@@ -109,15 +109,20 @@ func (p *Pipe) EachLine(process func(string, *strings.Builder)) *Pipe {
 			output := strings.Builder{}
 			process(scanner.Text(), &output)
 			if p.Error() != nil {
-				q.SetError(p.Error())
 				break
 			}
 			w.Write([]byte(output.String()))
 		}
-		err := scanner.Err()
+		// by this time, the previous pipe stage must have finished
+		err := p.Error()
 		if err != nil {
-			p.SetError(err)
 			q.SetError(err)
+		} else {
+			err := scanner.Err()
+			if err != nil {
+				p.SetError(err)
+				q.SetError(err)
+			}
 		}
 		w.Close()
 	}()
--- a/filters.go
+++ b/filters.go
@@ -444,3 +444,25 @@ func (p *Pipe) SHA256Sums() *Pipe {
 		out.WriteRune('\n')
 	})
 }
+
+// Stream lets the pipeline after it be executed in the streaming mode.
+// Like a Unix pipe, functions start simultaneously, each consuming the
+// output of its predecessor instantly. Function will stop when it detects
+// an error either in itself or from its predecessor. One can call
+// Synchronize() or any sink function to disable the streaming.
+func (p *Pipe) Stream() *Pipe {
+	if p == nil {
+		return p
+	}
+	return p.withAsync(true)
+}
+
+// Synchronize turns off streaming mode
+func (p *Pipe) Synchronize() *Pipe {
+	if p == nil || p.Error() != nil || !p.async {
+		return p
+	}
+	w := bytes.Buffer{}
+	io.Copy(&w, p.Reader)
+	return NewPipe().WithReader(bytes.NewReader(w.Bytes())).withAsync(false)
+}
--- a/filters.go
+++ b/filters.go
@@ -2,7 +2,6 @@ package script
 
 import (
 	"bufio"
-	"bytes"
 	"container/ring"
 	"crypto/sha256"
 	"encoding/hex"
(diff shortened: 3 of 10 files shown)


bitfield:
Hey @gty929 ! This looks interesting!

I'm still not sure what problem this is intended to solve, though. Perhaps you could give an example? (I presume that you don't actually need to write a script that prints `toc` at one-second intervals.)

gty929:
Hi, @bitfield ! I used the example above just to show that the streaming is working. I think there have been quite a few discussions in issues #34, #59, and #78 about the use cases, and why we should not leave out the streaming functionality. 

E.g., as @besi1z and @xxxserxxx mentioned, it would be great if the output of `Exec()` can be displayed in real-time. As for me, I hate it when a long-running command does not produce anything on the screen (since I have no idea when the program will finish or whether anything gets stuck).

The second reason is for time and memory efficiency, as @posener and @kepkin mentioned. By making the pipeline asynchronous, the output from one stage can be consumed by the next stage in time. For example, to find all the TODO's in 10,000 files, it's really not necessary to read all the files into memory.

By the way, I just rewrote `Exec()` so that it supports streaming as well. One can check it with besi1z's example: 
```go
script.Exec("ping -c 100 www.google.com").Stdout()
```

If the efforts are worthy, I will write more integration tests for the streaming feature. I guess it's not easy to fully test that streaming works as expected (i.e., write test cases that fail with the original code and pass with the new code) since the program is asynchronous. But I can write some tests to ensure that the library doesn't do anything bad, like producing a wrong output, crashing, or getting stuck forever.

bitfield:
> For example, to find all the TODO's in 10,000 files, it's really not necessary to read all the files into memory.

Indeed, and we don't! If you look at the example `grep` implementation, we have:

```go
	script.Stdin().Match(os.Args[1]).Stdout()
```

The `Match` filter reads one line at a time and checks it for a match, so no matter how much data is supplied, we never hold more than a line at a time in memory.

Regarding your other point, if you want to run a command like `ping -c 100 www.google.com`, that sounds like something that's best done by running the `ping` command directly. There's no value in simply executing that command from a Go program and sending its results to the standard output, so I don't think the ability to do that justifies the extra complexity that this PR would introduce.

gty929:
I noticed that there is already a "find TODO" example in the library. The program prints the filename and the line number before each line, so the implementation a bit more complex than the `grep` example.
```go
func main() {
	listPath := "."
	if len(os.Args) > 1 {
		listPath = os.Args[1]
	}
	// filter hidden directories and files
	filterFiles := regexp.MustCompile(`^\..*|/\.`)
	files := script.FindFiles(listPath).RejectRegexp(filterFiles)
	content := files.EachLine(func(filePath string, builderFile *strings.Builder) {
		p := script.File(filePath)
		lineNumber := 1
		p.EachLine(func(str string, build *strings.Builder) {
			if strings.Contains(str, "todo") {
				builderFile.WriteString(fmt.Sprintf("%s:%d %s \n", filePath, lineNumber, strings.TrimSpace(str)))
			}
			lineNumber++
		})
	})
	content.Stdout()
}
```
Even though the program does not read in the file contents all at once, it still accumulates all the outputs before sending them to Stdout, which I believe is not ideal. For example, a user may be running a search program like this on a large project or file system, and s/he may lose patience if no result gets printed for a long time. What's worse, if the user makes a mistake, e.g., writing a wrong regex expression in the `if` statement, s/he won't notice it until all the files are scanned. (In the worst case, if the expression matches every line, then there might be an out-of-memory problem.)
 
I just wrote another example for my proposed stream functionality. For multi-threaded programs, developers usually need to run a test suite many times to detect concurrency bugs. With the `-count` flag, `go test` can be run multiple times automatically. A problem, however, is that the command will not output anything unless it encounters a failure, so it's hard for developers to track the progress. (Adding a `-v` flag, on the other hand, will cause the program to output too much information.) The following example program solves this problem by outputting progress information after each round of testing.
```go
// This program runs tests on the script library 50 times and prints out the progress nicely.
func main() {
	round, step := 10, 5
	progressInfo := make([]string, round)
	for i := 0; i < round; i++ {
		progressInfo[i] = fmt.Sprintf("------ Done %v / %v ------", (i+1)*step, round*step)
	}
	cmd := fmt.Sprintf("bash -c 'go test -count %v github.com/bitfield/script; echo {{.}}'", step)
	// with Stream(), the program can print to stdout in real time
	script.Slice(progressInfo).Stream().ExecForEach(cmd).Stdout()
}
```
Here's a sample output, which is printed out line by line:
```
ok      github.com/bitfield/script      1.538s
------ Done 5 / 50 ------
ok      github.com/bitfield/script      1.479s
------ Done 10 / 50 ------
ok      github.com/bitfield/script      1.440s
------ Done 15 / 50 ------
...
```
Obviously, we cannot write a program like this in Script without the streaming feature.

I agree that we should avoid unnecessary complexity, but I do think that the task is worth the effort. As I mentioned above, there have been at least four other people who have demanded this feature or have tried to implement it on their own.

To make the program backward compatible, we just need to add two more public functions: a Stream() function that makes the subsequent pipe functions executed in the streaming mode, as demonstrated above, and a Synchronize() function, which stops the streaming. For clients, the learning cost for the streaming feature is pretty low (while they can also choose not to use it). In fact, if the client is familiar with Pipe in UNIX systems, s/he may find the streaming version more natural.

From the developers' perspective, the cost of this change is almost one-off.  One need not worry about streaming when writing source and sink functions. For filter functions, no change is necessary as well if it returns `p.EachLine(...)` directly. For the remaining few filters, contributors just need to remember to set the "streaming" flag if a new pipe is initiated.

bitfield:
Yes, this sounds good! Why would we need `Stream/Synchronize` functions, though? Why not make all pipes streaming by default?

gty929:
If we use streaming by default, there's no way to make everything backward compatible. 

First, by the definition of streaming, a function cannot perform a full error check on its previous stage before it starts executing. The best assurance one can obtain is that each stage will set the error field before it closes the writer of `io.Pipe()`, so that the error can get caught by the next stage, thus propagating to the end of the pipeline. (See the commit 886d0bb).
If the program is read-only (e.g., counting lines), it's not a big deal, but if the operation contains some sensitive write operation, a user may want to ensure that each stage executes correctly, before moving to the next stage. In that case, a Synchronous() method will be helpful.

Second, there's no way for a filter function to know whether it is the last stage in a pipeline. Ideally, a user should end a pipeline with a sink function (which always synchronizes the pipeline), but it's not always the case. For example, a user may call the `EachLine()` method just to update a variable in the program, and hence ignore the method's return value. If we set streaming as default, `EachLine()` will return immediately, and it's likely that the Golang variable hasn't been updated yet. On the other hand, such asynchronous behavior could be desirable in some cases, e.g, a user may want to split a long pipeline into several pieces, while keeping the streaming property. Therefore, I guess it's better to just leave the choice to the users. 

Finally, as  @posener mentioned in issue #34, the behavior of `Exec` and `Eachline` changes slightly if streaming is turned on. In `Exec`, we cannot use `CombinedOutput` for streaming as before. In my current implementation, stderr is appended after stdout in `io.MultiReader`. Therefore, the sequence of the output lines may change. (I think in Unix, only stdout is pipelined. It's indeed a bit strange to combine stdout and stderr together in a pipeline.)
For `Eachline`, the effect of resetting the string builder will also be different. In the non-streaming implementation, the whole output of the function will be cleared if a reset is called on the string builder. But that is not acceptable in the streaming scenario, since the previous outputs have already been sent to the next stage. As a result, only the output corresponding to the specific line will get cleared. 

I guess adding these two functions is not a heavy burden to the library. After all, it's better than creating a new project called 'script-async' or whatever. If a user really wants streaming to be set as default, s/he can fork the library and change the initial value of the asynchronous flag in the NewPipe() function.

posener:
@gty929 as for another project you suggests. I already created https://github.com/posener/script

bitfield:
Yes, I see what you mean. Streaming is fundamentally a different model from the 'error-safe reader' which `script` implements, as described in chapter 10 of [The Power of Go: Tools](https://bitfieldconsulting.com/books/tools). It doesn't fit with everything else, and you'd really need a new type: `StreamPipe`, or something like that, to avoid confusion with the existing semantics.

I think you've convinced me that the cost of adding this functionality would just be too high; closing accordingly, with thanks for the work you've put in to prove the concept.

@posener, it's quite up to you of course, but it might be worth thinking about giving your library a different name, since it is effectively a different library now, rather than a modified fork of this one.

PR #89, opened 2021-12-05
Bug fixes and improvements for corner cases

gty929:
When working on #88 , I noticed some defects in the current library, so I'd like to submit this separate patch, which does the following: 

**Fix the defect that Column() crashes on zero or a negative input**
The library panics when the following code is run
```go
script.Echo("1").Column(-1)
```
After the change, the program will return an empty pipe.

**Fix the defect that Last() crashes on a negative input**
The library panics when the following code is run
```go
script.Echo("1").Last(-1)
```
After the change, the program will return an empty pipe.

**Fix the defect that Regex filters crash on a nil regex expression**
The library panics when the following code is run
```go
script.Echo("1").MatchRegexp(nil)
```
After the change, the program will return a pipe with the error "nil regular expression"

**Add unit test for WithError(); close the pipe when a non-nil error is set**
When fixing the regex defect, I just wrote
```go
if re == nil { // to prevent SIGSEGV
	return p.WithError(errors.New("nil regular expression"))
}
```
Then, I realized that `p` will not be automatically closed, since the subsequent filters and sinks will catch the error and return immediately. To fix this problem and prevent contributors from making the same mistake in the future, I think the best approach is to just close the pipe reader when a non-nil error is set.

**Add ExecForEach, Reject(), and RejectRegexp to doMethodsOnPipe tests**
Just some jobs that previous contributors forgot.

**Improve coverage by adding equal counts in the test case of Freq**
To test that lines are sorted in alphabetical order when there is a tie.

**Use io.Copy() in Stdout() to improve efficiency**
It seems unnecessary to write the pipe's reader into a string and then print it.

By the way, after all the changes, the total line coverage has risen from 92.7% to 93.8%.

### DIFF
--- a/filters.go
+++ b/filters.go
@@ -6,6 +6,7 @@ import (
 	"container/ring"
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -41,7 +42,7 @@ func (p *Pipe) Basename() *Pipe {
 func (p *Pipe) Column(col int) *Pipe {
 	return p.EachLine(func(line string, out *strings.Builder) {
 		columns := strings.Fields(line)
-		if col <= len(columns) {
+		if col > 0 && col <= len(columns) {
 			out.WriteString(columns[col-1])
 			out.WriteRune('\n')
 		}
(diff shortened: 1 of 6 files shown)


gty929:
@bitfield Hi, would you please take a look at this pull request? Many thanks.

bitfield:
Excellent work @gty929, thanks a lot!

PR #91, opened 2021-12-08
使Exec函数返回的错误信息更可读

jlvihv:
通常，即使命令执行失败，命令的输出也会揭示失败的原因，将这种错误信息返回给开发者，以帮助他们更好的排查错误原因。

### DIFF
--- a/filters.go
+++ b/filters.go
@@ -6,6 +6,7 @@ import (
 	"container/ring"
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


bitfield:
Thanks, @jlvihv, but I think this is already handled by `os/exec` itself, by means of the [ExitError](https://pkg.go.dev/os/exec#ExitError) type. If the command can't be run, then there won't be any `CombinedOutput`. If it _can_ be run, but returned a non-zero exit status, then the error already contains the command's stderr output.

Issue #87, opened 2021-11-20
Parallel execution for ExecForEach and EachLine


gty929:
Hi, this issue follows the discussion in #86. Up to now, the script runs in a fully synchronous manner. While in #34 and #59 people have come up with some brilliant ideas in asynchronous pipeline streaming, the designs and implementations just seem too complicated, as @bitfield mentioned. 

Here, I want to suggest a compromise -- adding methods `EachLineConc()` and  `ExecForEachConc()`, which should have the same input and output interface as `EachLine()` and  `ExecForEach()`, but enable parallel execution within the method. (E.g., `script.Slice(make([]string, 5)).ExecForEachConc("sleep 1")` should return in about 1 second, rather than 5 seconds.)

The `ExecForEachConc()` method shares the same use cases with GNU parallel and the `&` symbol in bash, e.g. :
[https://blogs.sas.com/content/sgf/2021/04/14/using-shell-scripts-for-massively-parallel-processing/](url)
[https://unix.stackexchange.com/questions/103920/parallelize-a-bash-for-loop/103922](url)
[https://askubuntu.com/questions/431478/decompressing-multiple-files-at-once](url)
[https://superuser.com/questions/538164/how-many-instances-of-ffmpeg-commands-can-i-run-in-parallel/547340#547340](url)

Apart from improving efficiency, I sometimes want to run programs concurrently for testing purposes. For example, two weeks ago, I just wrote a file transmission service, and I would like to test whether the receiver application behaves correctly when multiple files are sent to it at the same time. It would be nice if I could run something like:
```script.ListFiles("FILE*.in").ExecForEachConc("./wSender --ip 10.0.0.1 --port 8888 --file {{.}}").Stdout()```

Note: it seems that we cannot directly append & to the commands in `ExecForEach()` for this purpose. 
`script.Slice(make([]string, 5)).ExecForEach("sleep 1 &")` returns instantly, because there is no 'wait'. 
Also, if I change the argument to "bash -c 'sleep 1' &" or "bash -c 'sleep 1 &'", the program will still run for 5 seconds.

**Comment by gty929** [comment]
The implementation won't be too hard. We just need to rewrite `EachLineConc()`, and let `ExecForEachConc()` call this new method.
Here's an implementation that I first think of (which can for sure be further optimized):
```
func (p *Pipe) EachLineConc(process func(string, *strings.Builder)) *Pipe {
	if p == nil || p.Error() != nil {
		return p
	}
[...]

PR #92, opened 2021-12-09
Implementation of parallel execution for ExecForEach() and EachLine()

gty929:
The pull request relates to issue #87 and implements the parallel execution for ExecForEach() and EachLine().

The code has been fully tested to ensure that the order of input is preserved (by executing a large test case 1000 times) and there is no race condition (by running `go test -race`).


bitfield:
Now that we have concurrent filters, would it make sense to reimplement this using them?

PR #101, opened 2022-03-21
Make pipe filters concurrent

bitfield:


gty929:
I'm glad that the filters finally become concurrent. While this change looks similar to #88, I'm curious what makes you decide to give up the property of "error-safe reader" and turn to concurrent programming. @bitfield

bitfield:
Well, we can still be somewhat error-safe, but the main aim of the `script` library is to get things done, without too much paperwork!

PR #108, opened 2022-04-07
Added `AppendFileUnique()` function

rew1nter:
Suppose we have two files as following:
```
$ cat a.file
a
a
1
1
b
```
```
$ cat b.file
a
2 
```
Running `script.Exec("cat a.file").AppendFileUnique("b.file")` will change the b.file into:
```
$ cat b.file
a
2
1
b
```

#### AppenFileUnique() 
* Takes the data from pipe, then uniquely appends it to the specified file. 
* It avoids appending the line if its already in the specified file. 
* Creates the specified file if it doesn't exist. 
* Returns the unique list of slice containing appended values

### DIFF
--- a/script.go
+++ b/script.go
@@ -897,3 +897,4 @@ func fileExists(filepath string) bool {
 	}
 	return false
 }
+
--- a/script.go
+++ b/script.go
@@ -5,6 +5,7 @@ import (
 	"container/ring"
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


bitfield:
There's quite a lot going on here! To help us get a clear picture of what the API should look like, could you give an example of a real-world program that would use this feature? Presumably you don't need to append unique lines like "a", "b" and so on to some file—what kind of lines *do* you need to append this way?

rew1nter:
as a security person, I enumerate IP addresses, company info, domains and subdomains etc. I use different tools and techniques to gather data but in fine I have to deduplicate them. go is my go-to scripting language so this is why I added it there.

bitfield:
Great! If you can rough out the program you want to write, then that'll be very helpful.

rew1nter:
I'm afraid not

bitfield:
No problem! We can always return to this later if you can come up with an example use case.

Issue #113, opened 2022-04-26
Add SHA-512 Hashing Sink


terrorbyte:
# Use Case

Currently SHA-256 is the only supported hashing algorithm (and there is some engineering design decisions in #39), this PR creates a SHA-512 sink that works identically the the current SHA-256 design but just with the `crypto/sha512` library imported and the same functions applied.

## Tests

As per the contributor guidelines I ran also created all the tests and test data. Here is the following output:

```console
$ go test -v | grep -i sha512
=== RUN   TestSHA512Sums_OutputsCorrectHashForEachSpecifiedFile
=== PAUSE TestSHA512Sums_OutputsCorrectHashForEachSpecifiedFile
=== RUN   TestSHA512Sum_OutputsCorrectHash
=== PAUSE TestSHA512Sum_OutputsCorrectHash
=== CONT  TestSHA512Sum_OutputsCorrectHash
=== RUN   TestSHA512Sum_OutputsCorrectHash/for_no_data
=== RUN   TestSHA512Sum_OutputsCorrectHash/for_short_string
=== RUN   TestSHA512Sum_OutputsCorrectHash/for_string_containing_newline
--- PASS: TestSHA512Sum_OutputsCorrectHash (0.00s)
    --- PASS: TestSHA512Sum_OutputsCorrectHash/for_no_data (0.00s)
    --- PASS: TestSHA512Sum_OutputsCorrectHash/for_short_string (0.00s)
    --- PASS: TestSHA512Sum_OutputsCorrectHash/for_string_containing_newline (0.00s)
=== CONT  TestSHA512Sums_OutputsCorrectHashForEachSpecifiedFile
--- PASS: TestSHA512Sums_OutputsCorrectHashForEachSpecifiedFile (0.00s)
=== RUN   ExamplePipe_SHA512Sum
--- PASS: ExamplePipe_SHA512Sum (0.00s)
=== RUN   ExamplePipe_SHA512Sums
--- PASS: ExamplePipe_SHA512Sums (0.00s)
```

One note, in the original patchset I added a `testdata/sha512Sum.input.txt` file that has the same content as `testdata/sha256Sum.input.txt`. This was to meet the contribution guidelines, but I think that it could be beneficial to generalize the test data to work with arbitrary hashing functions (ie something like `testdata/hashSum.input.txt`).

## Documentation

I updated the docs as well to match the SHA-256 functions. I am happy to add some more README examples for both the functions if need be.

terrorbyte:
[...]

PR #114, opened 2022-04-26
Added SHA-512 Hashing

terrorbyte:
> See #113 for discussion

Added a SHA-512 hashing mechanism modeled off of the SHA-256 tools. The majority of this code is nearly copy and pasted from the function with the SHA-512 crypto code. The tests have also been added and utilizes the SHA-512 `openssl dgst`, the only file added was a input file for SHA-512, though it is exactly the same as the 256 test file (I will bring up in the issue that it may be better to create just a generic file for hashing algorithms vs just for each one as per the nature of hashing algorithms).

All tests passed on my first run:

```
$ go test -v | grep -i sha512
=== RUN   TestSHA512Sums_OutputsCorrectHashForEachSpecifiedFile
=== PAUSE TestSHA512Sums_OutputsCorrectHashForEachSpecifiedFile
=== RUN   TestSHA512Sum_OutputsCorrectHash
=== PAUSE TestSHA512Sum_OutputsCorrectHash
=== CONT  TestSHA512Sum_OutputsCorrectHash
=== RUN   TestSHA512Sum_OutputsCorrectHash/for_no_data
=== RUN   TestSHA512Sum_OutputsCorrectHash/for_short_string
=== RUN   TestSHA512Sum_OutputsCorrectHash/for_string_containing_newline
--- PASS: TestSHA512Sum_OutputsCorrectHash (0.00s)
    --- PASS: TestSHA512Sum_OutputsCorrectHash/for_no_data (0.00s)
    --- PASS: TestSHA512Sum_OutputsCorrectHash/for_short_string (0.00s)
    --- PASS: TestSHA512Sum_OutputsCorrectHash/for_string_containing_newline (0.00s)
=== CONT  TestSHA512Sums_OutputsCorrectHashForEachSpecifiedFile
--- PASS: TestSHA512Sums_OutputsCorrectHashForEachSpecifiedFile (0.00s)
=== RUN   ExamplePipe_SHA512Sum
--- PASS: ExamplePipe_SHA512Sum (0.00s)
=== RUN   ExamplePipe_SHA512Sums
--- PASS: ExamplePipe_SHA512Sums (0.00s)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,7 @@ Sinks are methods that return some data from a pipe, ending the pipeline and ext
 | [`CountLines`](https://pkg.go.dev/github.com/bitfield/script#Pipe.CountLines) | |number of lines, error  |
 | [`Read`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Read) | given `[]byte` | bytes read, error  |
 | [`SHA256Sum`](https://pkg.go.dev/github.com/bitfield/script#Pipe.SHA256Sum) | | SHA-256 hash, error  |
+| [`SHA512Sum`](https://pkg.go.dev/github.com/bitfield/script#Pipe.SHA512Sum) | | SHA-512 hash, error  |
 | [`Slice`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Slice) | | data as `[]string`, error  |
 | [`Stdout`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Stdout) | standard output | bytes written, error  |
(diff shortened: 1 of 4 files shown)


bitfield:
Yes, this is great, but it seems a shame to duplicate all the SHA-256 code. Since both SHA-256 and SHA-512 implement `Hasher`, should we rewrite this to use common hashing code, but switch the hasher depending on the method called?

In other words, something like:

```go
func (p *Pipe) SHA256Sum() (string, error) {
	return p.hash(sha256.New())
}
```

bitfield:
I think this is superseded by #130, isn't it?

PR #119, opened 2022-05-17
Fix exit code issue

### LINKED ISSUE #117: ExitStatus of Exec
Hi,

what should be the exit code in the example below? I get 0 in both cases. I was expecting something other than 0.

```
package main
import (
	"fmt"

	"github.com/bitfield/script"
)

func main() {
	p := script.Exec("non-existant-command")
	fmt.Println(p.ExitStatus())

	p1 := script.Exec("false")
	fmt.Println(p1.ExitStatus())
}
```

thiagonache:
Add a non-existent command regex pattern and remove the close pipe from the setError function.

Closes #117

### DIFF
--- a/script.go
+++ b/script.go
@@ -103,6 +103,7 @@ func (p *Pipe) Error() error {
 }
 
 var exitStatusPattern = regexp.MustCompile(`exit status (\d+)$`)
+var nonExistentCmdPattern = regexp.MustCompile(`executable file not found in`)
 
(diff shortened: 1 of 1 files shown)


thiagonache:
Eg:
```go
package main

import (
	"fmt"

	"github.com/bitfield/script"
)

func main() {
	p := script.Exec("bogus")
	p.Wait()
	fmt.Println(p.ExitStatus())
}
```

PR #122, opened 2022-06-02
Ignore fs.WalkDir errors in FindFiles

### LINKED ISSUE #99: FindFiles exits and returns no files when it can't open a directory
This program

```
package main

import (
	"log"
	"os"

	"github.com/bitfield/script"
)

func main() {
	h, err := os.UserHomeDir()
	if err != nil {
		log.Fatal(err)
	}
	_, err = script.FindFiles(h).Stdout()
	if err != nil {
		log.Fatal(err)
	}
}
```

yields:

```
$ go build
$ ./script
2022/03/12 11:51:51 open /Users/user/.Trash: operation not permitted
```

I expected behavior similar to `find $HOME -type f`, i.e. to return files from directories it can open.

parkerduckworth:
Updates `FindFiles` to use `fs.WalkDir` instead of `filepath.Walk`, and ignores any errors that occur during the search.

Closes #99

### DIFF
--- a/script_test.go
+++ b/script_test.go
@@ -1017,6 +1017,25 @@ func TestFindFiles_RecursesIntoSubdirectories(t *testing.T) {
 	}
 }
 
+// FindFiles has been updated to ignore all errors
+// see [github issue #99]
+// (https://github.com/bitfield/script/issues/99)
+func TestFindFiles_FailsSilently(t *testing.T) {
+	t.Parallel()
+	want := "\n"
+	p := script.FindFiles("testdata/nonexistent")
+	if p.Error() != nil {
+		t.Fatal(p.Error())
+	}
+	got, err := p.String()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if want != got {
+		t.Error(cmp.Diff(want, got))
+	}
+}
+
 func TestIfExists_ProducesErrorPlusNoOutputForNonexistentFile(t *testing.T) {
 	t.Parallel()
 	want := ""


script_test.go
```diff
@@ -1017,14 +1017,6 @@ func TestFindFiles_RecursesIntoSubdirectories(t *testing.T) {
 	}
 }
 
-func TestFindFiles_InNonexistentPathReturnsError(t *testing.T) {
```
bitfield, line 1020: If it's a key part of `FindFiles` behaviour that it _doesn't_ return an error for missing paths, shouldn't we have a test for that?

parkerduckworth: Yeah great point, pushed up a test

script_test.go
```diff
@@ -1017,11 +1017,22 @@ func TestFindFiles_RecursesIntoSubdirectories(t *testing.T) {
...
+// FindFiles has been updated to ignore all errors
+// see [github issue #99]
+// (https://github.com/bitfield/script/issues/99)
+func TestFindFiles_FailsSilently(t *testing.T) {
```
bitfield, line 1023: I think "fails silently" could be misinterpreted, couldn't it? Like "fails" suggests it's not working properly. What about something like:

```suggestion
func TestFindFiles_ReturnsEmptyPipeAndNoErrorForNonexistentPath(t *testing.T) {
```

But I'm not convinced this is the right behaviour, actually. I *am* convinced that if there's some error from `fs.WalkDir`'s traversal, we should ignore that. But if the user has asked us to `FindFiles` starting from some _root_ that doesn't exist, I think that _should_ be an error. Would you want to know that, as the programmer? I would.

kwilczynski:
Hi @parkerduckworth,

This new functionality does make a lot of sense from a pure UX point of view. However, it might ever so slightly violate the [Principle of Least Surprise](http://principles-wiki.net/principles:principle_of_least_surprise).

Why? You might ask. A number of reasons come to mind, for instance:

- Hides errors...
- ... which changes the existing behaviour of this function
- Does not allow you to alter the behaviour, neither _enable_ nor _disable_ the new behaviour
- Changes the assumption about empty results: was there an error (and if so, then what was it?), or the files or directories are simply missing?

The change we are adding claims that this new functionality aligns **FindFiles()** closer to how **find** (man 1 find) works by ignoring errors. However, one could argue that find does not ignore errors per se - it does not stop execution when it can't **stat()** (the system call) a file or directory; however, the errors are very much not ignored, as errors are printed to the standard error output and should there be even a single I/O error, then find would return a non-zero exit code once it finishes to let you know that there was something wrong during its execution. Additionally, **find** can return an error without any results, should the provided path be inaccessible, for example (perhaps not the best example):

Screenshot from 2022-07-09 21-14-04

This is something that countless people often rely on in their shell scripts. So, there is an API, and there is an interface between find and the user, of sorts, there. I, for one, would like to know that there was an error somewhere.

What do you think?

Perhaps a way forward here would be such where **FindFiles()** would collect errors it saw, so to speak, and then return an error type that carries details of multiple errors (see the [hashicorp/go-multierror](https://github.com/hashicorp/go-multierror) package). This would allow for the **Error()** function on the _Pipe_ type (the fluent interface of Script that makes it so nice to use can also be a design challenge) to work as usual (since the **error** interface is still the one we return), and users would be able to reason about any potential errors.

A different approach would be to perhaps introduce a new function called **Find()**, which takes one argument that denotes the type of what to filter (akin to `-type f` or `type d' when using **find**), and two other arguments - an error and results callbacks. Or perhaps only two arguments - a type and a callback (similarly to the existing **Filter()** function).

@bitfield, your thoughts? @parkerduckworth yours?

Thank you!

Krzysztof

bitfield:
Maybe a sensible compromise here is that:

1. If `FindFiles` has _any_ results, there is no error
2. If `FindFiles` has _no_ results, but there was at least one error, then that error becomes the pipe's error status.

What do you think?

lazysegtree:
The issue was closed by another PR
- https://github.com/bitfield/script/issues/99

@bitfield Can you close this PR?

PR #129, opened 2022-08-12
fix countlines

yazanmonshed:
fix coutlines


bitfield:
Sorry @yazanmonshed, I'm missing the context for this—what's this change about?

PR #130, opened 2022-08-20
Generic HashSum and HashSums

wishdev:
This would replace #113/#114  and "depreciate" SHA256Sum and SHA256Sums by adding more generic functions called HashSum and HashSums.

One could call p.HashSum("SHA-256") or p.HashSums("SHA3-512") as opposed to adding specific functions for each hash algorithm desired.

This uses the crypto package and the official blessed algorithms within.

There are probably some extra tests (specifically passing an invalid hash name) and some docs to add - but I wanted to make sure I was on the right path here.....

SHA256Sum and SHA256Sums are replaced internally with calls to HashSum("SHA-256") and HashSums("SHA-256").

Finally, the internals are condensed and HashSum and HashSums use a common hash function to which they pass their io.Reader

### DIFF
--- a/script.go
+++ b/script.go
@@ -3,7 +3,7 @@ package script
 import (
 	"bufio"
 	"container/ring"
-	"crypto/sha256"
+	"crypto"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -20,6 +20,8 @@ import (
 
 	"bitbucket.org/creachadair/shell"
 	"github.com/itchyny/gojq"
+
+	_ "crypto/sha256"
 )
 
 // ReadAutoCloser represents a pipe source that will be automatically closed
@@ -716,25 +718,6 @@ func (p *Pipe) ReplaceRegexp(re *regexp.Regexp, replace string) *Pipe {
 	})
 }
 
-// SHA256Sums reads a list of file paths from the pipe, one per line, and
-// produces the hex-encoded SHA-256 hash of each file. Any files that cannot be
-// opened or read will be ignored.
-func (p *Pipe) SHA256Sums() *Pipe {
-	return p.FilterScan(func(line string, w io.Writer) {
-		f, err := os.Open(line)
-		if err != nil {
-			return // skip unopenable files
-		}
-		defer f.Close()
-		h := sha256.New()
-		_, err = io.Copy(h, f)
-		if err != nil {
-			return // skip unreadable files
-		}
-		fmt.Fprintln(w, hex.EncodeToString(h.Sum(nil)))
-	})
-}
-
 // AppendFile appends the contents of the pipe to the specified file, and
 // returns the number of bytes successfully written, or an error. If the file
 // does not exist, it is created.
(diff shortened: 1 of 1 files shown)


bitfield:
Thanks, @wishdev! This is certainly more flexible than what we have at the moment.

I'm not sure if specifying the desired hash algorithm as a string is the right way to go. It would make sense to define these as constants:

```go
script.Echo("hello").HashSum(script.SHA-256).Stdout()
```

However, this isn't valid Go, because identifiers can only contain letters. This is presumably why the constants defined in `crypto` are spelled the way they are: https://pkg.go.dev/crypto#Hash

It also sets up a potentially irksome maintainer chore of keeping the list up to date with Go's `crypto` library. So maybe we should just use _its_ constants:

```go
script.Echo("secret password").Hash(crypto.SHA256).Stdout()
```

wishdev:
Sounds perfectly reasonable. Updated...

bitfield:
Nice! Shall we now think about an example that demonstrates the need for `HashSums` with a non-trivial `script` pipeline?

wishdev:
'HashSums' isn't new functionality. It replaces SHA256Sums which is used for hashng files. You pass it a slice of file names and it will return the hash of each file. There is a test already setup for SHA256Sums within the test file.

bitfield:
> 'HashSums' isn't new functionality. It replaces SHA256Sums which is used for hashng files

Well, that's exactly the point of my question: _why_ replace it?

wishdev:
I guess I'm just confused - it "replaces" the function in the same way HashSum replaces SHA256Sum - it allows one to select a different hash algorithm for the task at hand.

Both SHA256Sum and SHA256Sums still exist - they are just single line functions to HashSum(crypto.SHA256) or HashSums(crypto.SHA256). 

I did move SHA256Sums down in the code base to get all the hashing functionality a little closer together - maybe that's leading to the confusion?

terrorbyte:
I also want to point out what I an API concern with using `crytpo.Hash` that I laid out in https://github.com/bitfield/script/issues/113#issuecomment-1111470432, which essentially means that we can use traditional hashes but not necessarily newer hash functions with parameters like `scrypt` and others. I still think that using a generic hash like @wishdev has laid out is far more flexible and the use case function would be almost exactly like here https://github.com/bitfield/script/issues/113#issuecomment-1111470374 but with `Hash(crypto.SHA512)` instead of my `SHA512Sums()`. I can attest to regularly utilizing different hash functions depending on their availability more than necessarily a decision to select one.

bitfield:
@wishdev what do you think of #215 — could you live with the suggested API?

wishdev:
I'm not a fan of the .new() splashed around code when you call the functions as opposed to passing a constant - however.....

1) I can wrap #215 and get what I like - i.e. I can just have a function that takes the constant - calls .new() and passes that along
2) The way I wrote it assumes all hash.Hash interfaces are created without any parameters to .new() - so #215 actually is cleaner in long term flexibility in that it takes the interface no matter how it's created. That's certainly a big plus!

bitfield:
Thanks, @wishdev - you've been very helpful in developing this feature. Looks like we've got something that everybody can work with.

PR #138, opened 2022-10-27
Addition of mktemp functionality

ioef:


bitfield:
Great, but I'm not sure what value this adds over `os.CreateTemp` and `os.TempDir`? And how does it fit into a pipeline? It doesn't use the contents of the pipe, nor provide a pipe itself. Can you give an example of a program where this would be useful?

ioef:
Hello.

Maybe i need to work upon this further and make it write the contents of the pipe to the specified newly created temporary file.

bitfield:
My suggestion would be to start by writing a program that both solves a real user problem and requires this feature. If it's hard to do that, then the feature may not actually be necessary. But if you can, then the program will go a long way towards showing what the API of the feature should be.

https://twitter.com/bitfield/status/1585219871347982336

PR #157, opened 2022-12-10
Adding "Must" panic sinks.

oderwat:
I added the sinks, but will not write test or change the documentation until I know that this is wanted and gets included.

This lets you write prototypes (or even production code) with less error checking cluttering.

For example:

```
goVersion := script.Get("https://go.dev/VERSION?m=text").MustString()
```
As panics() can be recovered, this can be an elegant strategy to write script like code, quite secure in my opinion.


oderwat:
retracted

PR #158, opened 2022-12-10
Fixing Go routine leaks because of not closing http requests.

oderwat:
We are using code that lets us detect dead locks which also checks for go routines that are still running when the program ends. I saw this getting triggered and found the cause in not telling the requests to close after use.

### DIFF
--- a/script.go
+++ b/script.go
@@ -590,6 +590,7 @@ func (p *Pipe) Get(URL string) *Pipe {
 	if err != nil {
 		return p.WithError(err)
 	}
+	req.Close = true
 	return p.Do(req)
 }
 
@@ -700,6 +701,7 @@ func (p *Pipe) Post(URL string) *Pipe {
 	if err != nil {
 		return p.WithError(err)
 	}
+	req.Close = true
 	return p.Do(req)
 }
 


bitfield:
Thanks, @oderwat! Just to help me understand the context of this, could you post the code you're using to detect goroutine leaks, the program you're analysing with it, and what the results are?

oderwat:
Sure:

```
mkdir leaktest
cd leaktest
go mod init leaktest
go get github.com/bitfield/script
go get github.com/ysmood/gotrace
```

Then create the file "leak_test.go" with the following code:

```
package main

import (
	"fmt"
	"github.com/bitfield/script"
	"testing"

	"github.com/ysmood/gotrace"
)

func TestHandle(t *testing.T) {
	// Just add one line before the standard test case
	gotrace.CheckLeak(t, 0)

	doit()
}

func doit() {
	fmt.Println(script.Get("https://go.dev/VERSION?m=text").String())
}
```

Run `go mod tidy`

Then run the test with:

```
GODEBUG="tracebackancestors=10" go test
```

You will see that your package will create a panic about the leaking go routine. This is fixed when you apply my changes.

```
$ GODEBUG="tracebackancestors=10" go test
go1.19.4 <nil>
--- FAIL: TestHandle (3.19s)
    leak_test.go:13: leaking goroutines: goroutine 22 [IO wait]:
        internal/poll.runtime_pollWait(0x9588bd0, 0x72)
                /usr/local/opt/go/libexec/src/runtime/netpoll.go:305 +0x89
```

Vs

```
$ GODEBUG="tracebackancestors=10" go test
go1.19.4 <nil>
PASS
ok      leaktest     0.306s
```

oderwat:
To be frank about it: The reuse of the connections with HTTP/1.1 keep alive is actually a feature. So one could argue that it is not a "bug" and that checking the still running Go routines would need to omit these routines. My problem with that is, that I use the check to catch "CTRL-C" on dead-loops so that I can see where it was stuck (or working) when the program gets canceled.

So from that perspective, one would need to ask if the "default" behavior of `script.Get()/script.Post()` should be the possible reuse of the connection. As far as I know, there is no way to tell the HTTP runtime to close all keep alive connections at the end of your program. It may work to make another request at the end and use the close flag on this.

oderwat:
I guess that `CloseIdleConnections()` on the client could also be used. So one could add `CloseIdle()` to the pipe, which lets one close the idle processes if one needs to do, what I need. I did not test it yet.

oderwat:
Yes. `CloseIdleConnections()` would work, but that needs to be called after the call is processes. So, it can't be done in the pipe, but one could add this to the sinks and create a Pipe, command that sets a flag "CloseIdle()". Another way would be to add variadic options to the Get() and Post() commands that let you add headers and also request closing the connection.

oderwat:
One could change the test like this:

```
package main

import (
	"fmt"
	"github.com/bitfield/script"
	"testing"

	"github.com/ysmood/gotrace"
)

func TestHandle(t *testing.T) {
	// Just add one line before the standard test case
	gotrace.CheckLeak(t, 0,
		gotrace.IgnoreCurrent(),
		gotrace.IgnoreFuncs("internal/poll.runtime_pollWait"))
	doit()
}

func doit() {
	fmt.Println(script.Get("https://go.dev/VERSION?m=text").String())
}
```

It would not report the open connection then. But it would also not report it :)

oderwat:
I will add a fix on out side.

PR #177, opened 2023-06-07
Apply operator: adding pre-composed operators together

skurfuerst:
Hey @bitfield ,

Thanks for this awesome toolbelt :)

I've felt one generic tool missing in here, basically a Transformer as explained here: https://blog.danlew.net/2015/03/02/dont-break-the-chain/

This allows code like the following, which allows for creation of higher level operators:

```go
	_, err := script.NewPipe().
		Apply(ExecAndStdout(`rm -Rf nsc/docs`)).
		Apply(ExecAndStdout(`mkdir nsc/docs`)).
		Stdout()


func ExecAndStdout(format string, args ...any) func(p *script.Pipe) *script.Pipe {
	return func(p *script.Pipe) *script.Pipe {
		p = p.Exec(fmt.Sprintf(format, args...)).Filter(StdoutAndContinue)
		p.Wait()
		// after p.Wait, the reader was fully consumed and auto-closing.
		// to prevent the error "io: read/write on closed pipe",
		// we need to re-initialize the reader.
		p.Reader = script.NewReadAutoCloser(strings.NewReader(""))
		return p
	}
}

```

If you generally give a +1 to the direction, I'll add detailed doc comments with examples like the above.

Thanks and all the best,
Sebastian

### DIFF
--- a/script.go
+++ b/script.go
@@ -193,7 +193,7 @@ func Stdin() *Pipe {
 }
 
 // Apply allows to 
-func (p *Pipe) Apply(fn func(p *Pipe) *Pipe) *.Pipe {
+func (p *Pipe) Apply(fn func(p *Pipe) *Pipe) *Pipe {
 	return fn(p)
 }
 
--- a/script.go
+++ b/script.go
@@ -192,6 +192,12 @@ func Stdin() *Pipe {
 	return NewPipe().WithReader(os.Stdin)
 }
 
+// Apply allows to 
+func (p *Pipe) Apply(fn func(p *Pipe) *Pipe) *.Pipe {
+	return fn(p)
+}
+
+
 // AppendFile appends the contents of the pipe to the file path, creating it if
 // necessary, and returns the number of bytes successfully written, or an
 // error.


bitfield:
Hi @skurfuerst! Thanks for the suggestion.

One thing I'm not totally clear on is exactly what problem this would solve. Can you elaborate on that a bit?

skurfuerst:
Hey @bitfield,

sure :) I read your book and especially the idea of "chaining" for CLI commands stuck totally with me to simplify error handling.

In my case, I have long lists of CLI commands which need to run after each other; and I wanted a mode like `set -ex` in the shell, where the full script aborts on the first error. By using `script`, I was not able to chain two Exec() calls after each other, and *still* displaying stdout for each of them.

Then I tried to pre-compose the `Exec` and `Stdout` operators together; though this did not fully work because `Stdout` is a kinda "terminal" operator (ending the chain); so I could not do "Exec().Stdout().Exec().Stdout().." (because of the function signatures).

Finally (the attempt displayed here) I looked at extending `script` by creating custom operators, to enjoy the benefits of the chaining - so that's why I added `Apply()` here.

Hope this gives some context :) Happy to answer any questions, and if you have other ideas how to solve these use cases, they'd be highly welcome.

Thanks and all the best,
Sebastian

bitfield:
Have you tried chaining the `Exec`s together directly? For example:

```go
script.Exec("cat foo").Exec("wc -l").Stdout()
```

skurfuerst:
Ah right, good point :) I tried this initially, and this has the problem that, because Exec is based on Filter, and Filter runs concurrently, the executed commands run concurrently - in my case, they have to run sequentially...

And I then tried to use .Exec().Wait(), but Wait is a terminal operation which is not chainable.

All the best and thanks for your work :) 
Sebastian

skurfuerst:
Ah right, good point :) I tried this initially, and this has the problem that, because Exec is based on Filter, and Filter runs concurrently, the executed commands run concurrently - in my case, they have to run sequentially...

And I then tried to use .Exec().Wait(), but Wait is a terminal operation which is not chainable.

All the best and thanks for your work :) 
Sebastian

bitfield:
That makes sense. The main use case for this library, though, is precisely to chain together operations in a pipeline, where the output of one stage becomes the input to the next.

You could run your sequenced operations like this, perhaps:

```go
p := script.Exec(“foo”)
p.Wait()
p.Exec(“bar”).Wait()
… // and so on
```

skurfuerst:
Yeah that would be an option, though I believe I would swallow std out then.

I think I will just go ahead and build my own library / abstraction then for my use cases, as for me these "multi pipes" / multiple exec steps are quite helpful / important. 

So no worries, shall we close this issue then? Up to you.

All the best,
Sebastian

bitfield:
Sure thing! Thanks for the discussion 😄

skurfuerst:
Thanks for your open source work and for your thoughts!

PR #179, opened 2023-07-13
Fix problem with ListFiles producing an unwanted Newline character when zero files are matched

### LINKED ISSUE #110: `ListFiles()` Produces an Unwanted Newline Character When Zero Files Are Matched
In the following example pipe, if the glob matches zero files, then a single empty line is supplied to `ExecForEach()`, which causes the `rm` command to error:

```
_, err := script.ListFiles("PY*").ExecForEach("rm -v {{.}}").Stdout()
```

Can `ListFiles()` only produce an output if files are matched?

There is a rather ugly workaround that filters blank lines as follows:

```
_, err := script.ListFiles("PY*").RejectRegexp(regexp.MustCompile("^$")).ExecForEach("rm -v {{.}}").Stdout()
```

MrTbag:
In this PR I tried to fix #110

Here I take into consideration whether any files are matched at all or not and then ListFiles() only produces an output if files are matched.

To fix the issue with the ListFiles function, you would need to handle the case where the provided path argument refers to a single file instead of a directory. Currently, the function assumes that if an error occurs while reading the directory, it's because the path matches exactly one file. However, this assumption is incorrect in some cases.

In the updated code:

-  We use os.Stat to check whether the path argument refers to a directory or a file.
- If it's a file, we return it as the output by creating a new Pipe and using Echo to set its content to the file path.
- If it's a directory, we proceed with the original logic of walking through the directory and appending file paths to the matches slice.

By making these changes, the ListFiles function will correctly handle both directory and file inputs.

#web_1402_issue

### DIFF
--- a/script.go
+++ b/script.go
@@ -132,31 +132,33 @@ func IfExists(path string) *Pipe {
 //
 // ListFiles does not recurse into subdirectories; use [FindFiles] instead.
 func ListFiles(path string) *Pipe {
-	if strings.ContainsAny(path, "[]^*?\\{}!") {
-		fileNames, err := filepath.Glob(path)
-		if err != nil {
-			return NewPipe().WithError(err)
-		}
-		return Slice(fileNames)
-	}
-	entries, err := os.ReadDir(path)
+	fileInfo, err := os.Stat(path)
 	if err != nil {
-		// Check for the case where the path matches exactly one file
-		s, err := os.Stat(path)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


bitfield:
Hi, @MrTbag! Thanks for the contribution—it's most welcome!

I'm not sure if I'm completely clear on the problem that this is fixing—could you help me out by supplying a couple of test cases that fail without this change?

PR #189, opened 2023-08-25
Implement Execv function (equivalent to Exec but with args passed as …

paulc:
Hello,

Really enjoying using `script` however there have been a couple of use cases where I have wanted to avoid interpolating the `Exec` args into a single cmdLIne (and worrying about quoting - which can be complex if dealing with arbitrary input) and have therefore added an `Execv(cmd string, args []string)` method which passes the cmd and args separately as []string (which is what gets passed to exec.Command anyway so saves re-parsing the command line).

Think this would be a useful addition (and be safer for scripts which make need to Exec commands where some of the parameters are untrusted)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ If you're already familiar with shell scripting and the Unix toolset, here is a 
 | Unix / shell       | `script` equivalent |
 | ------------------ | ------------------- |
 | (any program name) | [`Exec`](https://pkg.go.dev/github.com/bitfield/script#Exec) |
+|                    | [`Execv`](https://pkg.go.dev/github.com/bitfield/script#Execv) |
 | `[ -f FILE ]`      | [`IfExists`](https://pkg.go.dev/github.com/bitfield/script#IfExists) |
 | `>`                | [`WriteFile`](https://pkg.go.dev/github.com/bitfield/script#Pipe.WriteFile) |
 | `>>`               | [`AppendFile`](https://pkg.go.dev/github.com/bitfield/script#Pipe.AppendFile) |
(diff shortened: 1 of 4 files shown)


script.go
```diff
@@ -397,6 +404,33 @@ func (p *Pipe) Exec(cmdLine string) *Pipe {
...
+	return p.Filter(func(r io.Reader, w io.Writer) error {
+		cmd := exec.Command(cmd, args...)
+		cmd.Stdin = r
+		cmd.Stdout = w
+		cmd.Stderr = w
+		if p.stderr != nil {
+			cmd.Stderr = p.stderr
+		}
+		err := cmd.Start()
+		if err != nil {
```
isaqueveras, line 426:
```suggestion
		if err := cmd.Start(); err != nil {
```

jesselang:
I would very much appreciate having this slice-based Exec! Is this PR being considered?

README.md
```diff
@@ -177,6 +178,15 @@ PING 127.0.0.1 (127.0.0.1): 56 data bytes
 ...
 ```
 
+If we are construcing Exec arguments dynamically it can be easier to use Execv where we can pass cmd and []args separately without having to escape args.
+
+```go
+args := []string{};
+args = append(args,"127.0.0.1")
+script.Exec("ping",args).Stdout()
```
isaqueveras, line 186: Would that be `Execv`?
```suggestion
script.Execv("ping", args).Stdout()
```

lazysegtree:
@jesselang @paulc Are you still working on this ? Are you up for discussion regarding the feature?


@bitfield About the feature, are you okay with it? 
It makes certain use cases easier, but it will pollute the api. The we would also want to update `ExecForEach`. I am not sure if we should do this.

This is a similar PR. It seems like we dont want to implement it. I dont have a strong opinion either.
- https://github.com/bitfield/script/pull/200

bitfield:
I don't see any value to duplicating `exec.Command` here, no.

PR #200, opened 2024-05-02
Allow for script.Exec() to be used like exec.Command()

bartdeboer:
A small update to allow for `exec.Command(name string, arg ...string)` style arguments.

```go
script.Exec("go", "help")
```

### DIFF
--- a/script.go
+++ b/script.go
@@ -59,8 +59,8 @@ func Echo(s string) *Pipe {
 //
 // Use [Pipe.Exec] to send the contents of an existing pipe to the command's
 // standard input.
-func Exec(cmdLine string) *Pipe {
-	return NewPipe().Exec(cmdLine)
+func Exec(cmdLine string, args ...string) *Pipe {
+	return NewPipe().Exec(cmdLine, args...)
 }
 
 // File creates a pipe that reads from the file path.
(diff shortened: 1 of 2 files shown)


bitfield:
Hi @bartdeboer! Thanks for the PR.

Something that would be useful for documentation here is an example program showing where this feature would be used to solve a real-world problem. Can we come up with one?

bartdeboer:
A real world problem might be someone expecting the signature to work like `exec.Command`. I was now needing to look into the source and dependencies to find out what package was used to parse the command and see how it works with quoted arguments (arguments with spaces) etc. With this update people can use it either way they find most convenient.

bartdeboer:
Unfortunately I don't have an example program right now.

All I know is that `exec.Command` is widely used (also by script itself) and has served me well with complex commands. The shell.Fields method introduces a new approach to handling command lines unknown to me. Considering executing commands is such an important feature I think it will be nice to allow for such enhanced features while keeping the fallback on Go's core Exec functionality.

bitfield:
As a user, I appreciate simple APIs, and simplicity doesn't happen by accident. It happens by imposing a high bar that new features have to clear, and one of those should certainly be "solves a problem that someone has". 

To put it another way, if there's no problem, then there's no need for a PR! Let's keep it open, though, and this discussion might evolve to generate some use cases in the future.

lazysegtree:
@bartdeboer Are you still up for discussion about this, or we should close this PR?

PR #208, opened 2024-08-15
feat: add `WithEnv` method for setting a task's environment variables

### LINKED ISSUE #80: Nice way to handle env variable?
Some methods to set environment varialbes for single command, instead of this:

```go
os.Setenv("XB_AUTHOR_DATE", dateStr)
os.Setenv("XB_COMMITTER_DATE", dateStr)
```

Maybe in old shell way:

```bash
XB_AUTHOR_DATE="2021-04-05" XB_COMMITTER_DATE="2021-04-05" xb_push xxxx
```

mahadzaryab1:
Added a method `WithEnv` that takes in a string array (`[]string `) and overrides the environment executing a `exec.Command` to that array. This PR fixes #80.

### DIFF
--- a/script.go
+++ b/script.go
@@ -35,7 +35,7 @@ type Pipe struct {
 	mu  *sync.Mutex
 	err error
 
-	// env contains the environment to run the exec command with.
+	// env contains the environment to run any exec commands with.
 	// Each entry in the array should be of the form key=value.
 	// If env is not nil, it will replace the default environment variables
 	// when executing commands.
--- a/script.go
+++ b/script.go
@@ -32,10 +32,6 @@ type Pipe struct {
 	stdout     io.Writer
 	httpClient *http.Client
 
-	// because pipe stages are concurrent, protect 'err'
-	mu  *sync.Mutex
-	err error
-
 	// env contains the environment to run any exec commands with.
 	// Each entry in the array should be of the form key=value.
 	// If env is not nil, it will replace the default environment variables
--- a/script.go
+++ b/script.go
@@ -39,13 +39,7 @@ type Pipe struct {
 	mu     *sync.Mutex
 	err    error
 	stderr io.Writer
-	// env contains the environment to run any exec commands with.
-	// Each entry in the array should be of the form key=value.
-	// If env is not nil, it will replace the default environment variables
-	// when executing commands.
-	// If env is an empty array, any exec commands will be run with an
-	// empty environment.
-	env []string
+	env    []string
 }
 
 // Args creates a pipe containing the program's command-line arguments from
@@ -929,8 +923,9 @@ func (p *Pipe) Wait() error {
 
 // WithEnv sets the environment for subsequent [Pipe.Exec] and [Pipe.ExecForEach] commands
 // to the string slice env. This will override the default process environment variables
-// when executing commands run via [Pipe.Exec] or [Pipe.ExecForEach]. This will not affect
-// the current process's environment.
+// when executing commands run via [Pipe.Exec] or [Pipe.ExecForEach].
+// Each entry in the array should be of the form key=value.
+// If env is an empty array, any exec commands will be run with an empty environment.
 func (p *Pipe) WithEnv(env []string) *Pipe {
 	p.mu.Lock()
 	defer p.mu.Unlock()
--- a/script.go
+++ b/script.go
@@ -911,10 +911,10 @@ func (p *Pipe) WithStdout(w io.Writer) *Pipe {
 	return p
 }
 
-// WithEnv sets the environment for the exec commands to the string array env.
-// This will override the default process environment variables when executing
-// commands run via [Pipe.Exec] or [Pipe.ExecForEach]. This will not affect the
-// environment outside of [Pipe.Exec] or [Pipe.ExecForEach].
+// WithEnv sets the environment for subsequent [Pipe.Exec] and [Pipe.ExecForEach] commands
+// to the string slice env. This will override the default process environment variables
+// when executing commands run via [Pipe.Exec] or [Pipe.ExecForEach]. This will not affect
+// the environment outside of [Pipe.Exec] or [Pipe.ExecForEach].
 func (p *Pipe) WithEnv(env []string) *Pipe {
 	p.env = env
 	return p
--- a/script.go
+++ b/script.go
@@ -34,6 +34,12 @@ type Pipe struct {
 	// because pipe stages are concurrent, protect 'err'
 	mu  *sync.Mutex
 	err error
+
+	// env contains the environment to run the exec command with.
+	// Each entry in the array should be of the form key=value.
+	// If env is not nil, it will replace the default environment variables
+	// when executing commands.
+	env []string
 }
 
 // Args creates a pipe containing the program's command-line arguments from
@@ -166,6 +172,7 @@ func NewPipe() *Pipe {
 		mu:         new(sync.Mutex),
 		stdout:     os.Stdout,
 		httpClient: http.DefaultClient,
+		env:        nil,
 	}
 }
 
@@ -388,6 +395,9 @@ func (p *Pipe) Exec(cmdLine string) *Pipe {
 		if p.stderr != nil {
 			cmd.Stderr = p.stderr
 		}
+		if p.env != nil {
+			cmd.Env = p.env
+		}
 		err = cmd.Start()
 		if err != nil {
 			fmt.Fprintln(cmd.Stderr, err)
@@ -427,6 +437,9 @@ func (p *Pipe) ExecForEach(cmdLine string) *Pipe {
 			cmd.Stderr = w
 			if p.stderr != nil {
 				cmd.Stderr = p.stderr
+			}
+			if p.env != nil {
+				cmd.Env = p.env
 			}
 			err = cmd.Start()
 			if err != nil {
@@ -895,6 +908,14 @@ func (p *Pipe) WithStderr(w io.Writer) *Pipe {
 // default [os.Stdout].
 func (p *Pipe) WithStdout(w io.Writer) *Pipe {
 	p.stdout = w
+	return p
+}
+
+// WithEnv sets the pipe's environment to the string array env. This will override
+// the default process environment variables when executing commands run via [Pipe.Exec]
+// or [Pipe.ExecForEach].
+func (p *Pipe) WithEnv(env []string) *Pipe {
+	p.env = env
 	return p
 }
 
--- a/script.go
+++ b/script.go
@@ -32,16 +32,20 @@ type Pipe struct {
 	stdout     io.Writer
 	httpClient *http.Client
 
+	// mu protects the following fields because pipe stages are concurrent
+	// 	- err
+	// 	- stderr
+	// 	- env
+	mu     *sync.Mutex
+	err    error
+	stderr io.Writer
 	// env contains the environment to run any exec commands with.
 	// Each entry in the array should be of the form key=value.
 	// If env is not nil, it will replace the default environment variables
 	// when executing commands.
+	// If env is an empty array, any exec commands will be run with an
+	// empty environment.
 	env []string
-
-	// because pipe stages are concurrent, protect 'err' and 'stderr'
-	mu     *sync.Mutex
-	err    error
-	stderr io.Writer
 }
 
 // Args creates a pipe containing the program's command-line arguments from
@@ -381,6 +385,12 @@ func (p *Pipe) EncodeBase64() *Pipe {
 	})
 }
 
+func (p *Pipe) environment() []string {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	return p.env
+}
+
 // Error returns any error present on the pipe, or nil otherwise.
 // Error is not a sink and does not wait until the pipe reaches
 // completion. To wait for completion before returning the error,
@@ -426,8 +436,9 @@ func (p *Pipe) Exec(cmdLine string) *Pipe {
 		if pipeStderr != nil {
 			cmd.Stderr = pipeStderr
 		}
-		if p.env != nil {
-			cmd.Env = p.env
+		pipeEnv := p.environment()
+		if pipeEnv != nil {
+			cmd.Env = pipeEnv
 		}
 		err = cmd.Start()
 		if err != nil {
@@ -919,8 +930,10 @@ func (p *Pipe) Wait() error {
 // WithEnv sets the environment for subsequent [Pipe.Exec] and [Pipe.ExecForEach] commands
 // to the string slice env. This will override the default process environment variables
 // when executing commands run via [Pipe.Exec] or [Pipe.ExecForEach]. This will not affect
-// the environment outside of [Pipe.Exec] or [Pipe.ExecForEach].
+// the current process's environment.
 func (p *Pipe) WithEnv(env []string) *Pipe {
+	p.mu.Lock()
+	defer p.mu.Unlock()
 	p.env = env
 	return p
 }
--- a/script.go
+++ b/script.go
@@ -32,10 +32,6 @@ type Pipe struct {
 	stdout     io.Writer
 	httpClient *http.Client
 
-	// mu protects the following fields because pipe stages are concurrent
-	// 	- err
-	// 	- stderr
-	// 	- env
 	mu     *sync.Mutex
 	err    error
 	stderr io.Writer
@@ -403,6 +399,11 @@ func (p *Pipe) Error() error {
 // error output). The effect of this is to filter the contents of the pipe
 // through the external command.
 //
+// # Environment
+//
+// The command inherits the current process's environment, optionally modified
+// by [Pipe.WithEnv].
+//
 // # Error handling
 //
 // If the command had a non-zero exit status, the pipe's error status will also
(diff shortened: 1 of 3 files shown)


script_test.go
```diff
@@ -1768,6 +1769,57 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
 	}
 }
 
+func TestWithEnv_SetEmptyEnvironment(t *testing.T) {
```
bitfield, line 1772: Could we be a bit more descriptive here?

```suggestion
func TestWithEnv_UnsetsAllEnvVarsGivenEmptySlice(t *testing.T) {
```

script_test.go
```diff
@@ -1768,6 +1769,57 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
...
+	t.Parallel()
+	buf := new(bytes.Buffer)
+	env := []string{}
+
+	_, err := script.NewPipe().WithStdout(buf).WithEnv(env).Exec("printenv").Stdout()
```
bitfield, line 1777: If you're testing the output of some command, wouldn't it be slightly easier to call `String` on the pipe, rather than creating a buffer and using `WithStdout`?

script_test.go
```diff
@@ -1768,6 +1769,57 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(buf.String()) != 0 {
+		got := strings.Split(strings.TrimRight(buf.String(), "\n"), "\n")
+		t.Errorf("expected 0 environment variables, got %d %v", len(got), got)
```
bitfield, line 1784: I'm not sure the string-munging is all that helpful here, and clutters up the test a bit, making it hard to see what's going on. The point you're making is that the string should be empty. If it's not, we want to know what's in it. That's all.

```suggestion
		if output != "" {
		        t.Errorf("want empty environment, got %q", output)
		 }
```

script_test.go
```diff
@@ -1768,6 +1769,57 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
...
+	}
+}
+
+func TestWithEnv_SetMultipleEnvVars(t *testing.T) {
+	t.Parallel()
+	buf := new(bytes.Buffer)
+	env := []string{"ENV1=test1", "ENV2=test2"}
+
+	_, err := script.NewPipe().WithStdout(buf).WithEnv(env).Exec("printenv").Stdout()
```
bitfield, line 1793: It seems a shame to add an implicit dependency on the `printenv` command, doesn't it? Could we use the shell instead?

script_test.go
```diff
@@ -1768,6 +1769,57 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	got := strings.Split(strings.TrimRight(buf.String(), "\n"), "\n")
```
bitfield, line 1798: I think it's probably a bad idea to munge the output in this way, and it means we're not comparing against the _actual_ output of the command, we're testing against some arbitrarily-munged version. For tests to be clear and direct, it helps to specify the exact, literal string you expect to get.

script_test.go
```diff
@@ -1768,6 +1769,57 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
...
+	t.Parallel()
+	buf := new(bytes.Buffer)
+
+	_, err := script.NewPipe().WithStdout(buf).Exec("printenv").Stdout()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// not setting the environment should simply use the task's default environment
+	expected := os.Environ()
```
bitfield, line 1815: Just a house style issue, but we universally use the names `want` and `got` in tests, and the failure message "want %q, got %q" (for example).

bitfield (changes requested):
Great work so far! Just a couple of points:

* Do these tests work on Windows? (You can add Windows-specific tests to `script_windows_test.go`)
* What potential bugs in the environment-setting code are we failing to catch in tests here?

script_test.go
mahadzaryab1: @bitfield would you please be able to elaborate on this for me? i'm not sure what it means to use the shell instead

mahadzaryab1:
> Great work so far! Just a couple of points:
> 
> * Do these tests work on Windows? (You can add Windows-specific tests to `script_windows_test.go`)
> * What potential bugs in the environment-setting code are we failing to catch in tests here?

Thank you so much for the great feedback @bitfield! I addressed some of your feedback and had a couple of follow-up questions: 
- Is it possible to run the windows-specific tests locally? (I'm on my MacBook)
- For potential bugs - one possibility could be passing in malformed environment variables. Should we add some tests for this? Let me know if there's something else that I'm missing.

script_test.go
bitfield: Well, you'll notice that `script` itself doesn't rely on _any_ external commands being present—which is a good thing, because we basically can't rely on that! And the tests don't use many external commands either: currently just `go` (which seems at least reasonable) and `echo` (which exists most places, is usually a shell builtin, and even works on Windows, though with slightly different syntax).

If we could stick to those, without adding a new dependency that we don't absolutely need, that would be a nice bonus.

bitfield:
> * Is it possible to run the windows-specific tests locally? (I'm on a MacBook)

That's a good question—I'd like to be able to do this too. I can run them in CI, but of course it would be helpful for you to get the results locally. What about running the tests in a Windows Docker image? Would that work?

> * For potential bugs - one possibility could be passing in malformed environment variables. Should we add some tests for this? Let me know if there's something else that I'm missing.

Well, that's really the user's problem! We don't do anything but pass on what they gave us to the `exec.Cmd`, which is absolutely right. The kind of bugs I meant are the ones in *our* logic. Basically, we could forget to set `p.Env`, and we could forget to reference it in either `p.Exec` or `p.ExecForEach`, and I think the existing tests cover those. So we're good.

There is still the awkward question of what happens if someone writes some arbitrary `Filter` function on the pipe that does something like `os.Getenv`. You could imagine users thinking that `p.WithEnv` should affect this, but it won't. As far as I can see there's no way around this, unless _we_ called `os.Setenv`, which is exactly what we don't want to do. Maybe we could clarify this in the documentation somehow?

script_test.go
mahadzaryab1: @bitfield there is the `env` command but i'm not sure that's too different from using `printenv`. the other option would be to use `echo $ENV1` to capture the value of a single environment variable but i don't believe there's a way to print all the variables using `echo`. let me know what you think!

mahadzaryab1:
@bitfield do you know if the windows tests run in WSL?

bitfield:
> @bitfield do you know if the windows tests run in WSL?

I don't.

script.go
```diff
@@ -898,6 +911,14 @@ func (p *Pipe) WithStdout(w io.Writer) *Pipe {
 	return p
 }
 
+// WithEnv sets the pipe's environment to the string array env. This will override
```
lambrospetrou, line 914: One small suggestion, which relates to the discussion on the main page too, is that this doesn't modify the "pipe" environment, but the "next exec command's" environment. If thie documentation mentions this I believe it's clear and won't be confusing to folks accessing the env outside of the `Exec` functions.

mahadzaryab1: @lambrospetrou thanks for the suggestion! i clarified this comment now to add more detail about the usage. let me know what you think and if you have any other feedback

script_test.go
bitfield: If we could avoid adding a new external dependency, that would be great!

script.go
```diff
@@ -34,6 +34,12 @@ type Pipe struct {
 	// because pipe stages are concurrent, protect 'err'
 	mu  *sync.Mutex
 	err error
+
+	// env contains the environment to run any exec commands with.
+	// Each entry in the array should be of the form key=value.
+	// If env is not nil, it will replace the default environment variables
+	// when executing commands.
+	env []string
```
bitfield, line 42: Is this protected by the mutex? Does it need to be?

script.go
```diff
@@ -898,6 +911,15 @@ func (p *Pipe) WithStdout(w io.Writer) *Pipe {
 	return p
 }
 
+// WithEnv sets the environment for the exec commands to the string array env.
```
bitfield, line 914:
```suggestion
// WithEnv sets the environment for subsequent [Pipe.Exec] and [Pipe.ExecForEach] commands to the string slice env.
```

script_test.go
```diff
@@ -1768,6 +1768,52 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if output != "" {
+		t.Errorf("want empty environment, got %q", output)
+	}
+}
+
+func TestWithEnv_SetMultipleEnvVars(t *testing.T) {
```
bitfield, line 1785: Can we make this a sentence describing the behaviour?

bitfield (changes requested):

script_test.go
mahadzaryab1: @bitfield using `echo` doesn't actually work since the variables aren't exported. I'm not sure if there's another way to print the variables we want without using `env` or `printenv`.

script.go
mahadzaryab1: this should be protected by the mutex based on the race condition that https://github.com/bitfield/script/pull/209 fixes! i'll make the change

script.go
```diff
@@ -34,6 +34,12 @@ type Pipe struct {
 	// because pipe stages are concurrent, protect 'err'
 	mu  *sync.Mutex
 	err error
+
+	// env contains the environment to run any exec commands with.
+	// Each entry in the array should be of the form key=value.
+	// If env is not nil, it will replace the default environment variables
+	// when executing commands.
```
bitfield, line 41: I think it's worth pointing out also that if `env` is an _empty_ array, it will remove everything, and commands will be run with a completely empty environment. Since `nil` and empty arrays usually have the same semantics, this would otherwise be slightly surprising (but completely correct) behaviour.

script.go
```diff
@@ -858,6 +871,15 @@ func (p *Pipe) Wait() {
 	}
 }
 
+// WithEnv sets the environment for subsequent [Pipe.Exec] and [Pipe.ExecForEach] commands
+// to the string slice env. This will override the default process environment variables
+// when executing commands run via [Pipe.Exec] or [Pipe.ExecForEach]. This will not affect
+// the environment outside of [Pipe.Exec] or [Pipe.ExecForEach].
```
bitfield, line 877: "The environment outside" is a bit vague—shall we say explicitly that it doesn't affect the current _process's_ environment, in the way that `os.Setenv` would? (I can imagine people getting confused that this doesn't behave the way `Setenv` does.)

script.go
```diff
@@ -858,6 +871,15 @@ func (p *Pipe) Wait() {
...
+func (p *Pipe) WithEnv(env []string) *Pipe {
+	p.env = env
```
bitfield, line 879: We also need the mutex here.

script_test.go
```diff
@@ -1768,6 +1768,52 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if output != "" {
+		t.Errorf("want empty environment, got %q", output)
+	}
+}
+
+func TestWithEnvChangesExecutionEnvironment(t *testing.T) {
```
bitfield, line 1785:
```suggestion
func TestWithEnv_SetsGivenVariablesForSubsequentExec(t *testing.T) {
```

In case you didn't know, which by the way would be absolutely excusable, the `_` is a hint to [gotestdox](https://github.com/bitfield/gotestdox). It's a tool for printing out your test names as descriptive sentences (try it in this repo!), but without the `_` it doesn't know that `WithEnv` is a camel-cased function name and not just the words "with env".

script_test.go
```diff
@@ -1768,6 +1768,52 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
...
+	}
+
+	want := "ENV1=test1\nENV2=test2\n"
+
+	if got != want {
+		t.Errorf("want %v, got %v", want, got)
+	}
+}
+
+func TestNotSettingEnvFallsBackToDefaultEnvironment(t *testing.T) {
```
bitfield, line 1801: I don't honestly think we need this test (and that's not something I often find myself saying!)

After all, it's essentially saying that if you don't use `WithEnv`, nothing special happens—and that was already the case.

bitfield (changes requested):
Nice work so far!

script_test.go
bitfield: Yes, this is a bit of a subtle one. If you did something like this:

```go
Exec("echo ENV1=$ENV1")
```

that wouldn't work, because `$`-expansion is done by the _shell_, and we're not executing a shell! We're executing the `echo` command directly, which doesn't interpolate variables.

On the other hand, if we exec a shell instead:

```go
Exec("sh -c 'echo ENV1=$ENV1'")
```

this works as expected, because the shell expands `$ENV1` before passing it to `echo`.

We do use `sh -c` in a couple of existing tests, so there's no problem with depending on it here.

mahadzaryab1: That's really cool! Thanks for the explanation @bitfield

mahadzaryab1:
@bitfield addressed your feedback! What should we do about the `TestWithEnv_UnsetsAllEnvVarsGivenEmptySlice` test case? It currently relies on the printenv command and is failing on Windows.

bitfield:
> @bitfield addressed your feedback! What should we do about the `TestWithEnv_UnsetsAllEnvVarsGivenEmptySlice` test case? It currently relies on the printenv command and is failing on Windows. We can't echo a specific environment variable in this case since we're not setting anything.

Hmm, what about echoing some variable which would usually be there (`HOME` or `PATH`, for example), and checking that it's empty? That proves that `WithEnv` erased the environment.

We'll need a different one on Windows (`HOMEPATH`)?

script.go
bitfield: The place to say this, of course, would be in the doc comment for `WithEnv`—in fact, we don't really need to say anything much about the struct field itself.

mahadzaryab1:
> // env contains the environment to run any exec commands with.
> 	// Each entry in the array should be of the form key=value.
> 	// If env is not nil, it will replace the default environment variables
> 	// when executing commands.

@bitfield I tried a different approach where I do an `os.SetEnv` and test that it doesn't show up. when passing in an empty slice. What do you think of this approach? This was mostly to keep things platform-agnostic so we don't have to write different tests for different platforms.

script_test.go
```diff
@@ -1768,6 +1768,44 @@ func TestWithStdout_SetsSpecifiedWriterAsStdout(t *testing.T) {
 	}
 }
 
+func TestWithEnv_UnsetsAllEnvVarsGivenEmptySlice(t *testing.T) {
+	t.Parallel()
+	os.Setenv("ENV1", "test1")
```
bitfield, line 1773: Not a bad idea in principle, but unfortunately `Setenv` isn't safe to use in a parallel test, because it affects the environment for all tests. (If we _were_ going to do that, we could use `t.Setenv`, which does the cleanup automatically, but it's still not parallelisable.)

Actually, what we care about is not the whole _test binary's_ environment, but just the _pipe's_, isn't it? When you frame it that way, all we need to do is create a pipe with some environment variable already set, and check that `WithEnv` unsets it. Something like this, perhaps?

```go
func TestWithEnv_UnsetsAllEnvVarsGivenEmptySlice(t *testing.T) {
	t.Parallel()
	p := script.NewPipe().WithEnv([]string{"ENV1=test1"}).Exec("sh -c 'echo ENV1=$ENV1'")
	want := "ENV1=test1\n"
	got, err := p.String()
	if err != nil {
		t.Fatal(err)
	}
	if got != want {
		t.Fatalf("want %q, got %q", want, got)
	}
	got, err = p.Echo("").WithEnv([]string{}).Exec("sh -c 'echo ENV1=$ENV1'").String()
	if err != nil {
		t.Fatal(err)
	}
	want = "ENV1=\n"
	if got != want {
		t.Errorf("want %q, got %q", want, got)
	}
}
```

bitfield (changes requested):

script_test.go
mahadzaryab1: Great idea! Thanks @bitfield

bitfield (approved):
Great job, @mahadzaryab1, thanks a lot! Everybody wants to _propose_ features, and even do quick-and-dirty implementations of them, but that's easy. Few people are willing to do the slow, detailed, careful, high-quality implementation and polish that characterises durable software. You've done that on several issues, and that makes you `script`'s most valuable contributor by far!

Anything else you'd like to work on? I think we have enough to put a release together now, but if you could pick an existing issue to solve, or even suggest a new one, what would it be?

mahadzaryab1:
> Great job, @mahadzaryab1, thanks a lot! Everybody wants to _propose_ features, and even do quick-and-dirty implementations of them, but that's easy. Few people are willing to do the slow, detailed, careful, high-quality implementation and polish that characterises durable software. You've done that on several issues, and that makes you `script`'s most valuable contributor by far!
> 
> Anything else you'd like to work on? I think we have enough to put a release together now, but if you could pick an existing issue to solve, or even suggest a new one, what would it be?

@bitfield Thank you so much for your kind words and guidance in helping me working on these issues. I love writing Go and learnt a lot about the language from your code reviews! In terms of proposing new features, I would love to and I'll definitely start to give this more thought and open issues for any ideas that I have. However, in the meantime, I also do want to keep contributing to `script` so I wanted to ask if there were any features needed implementing? I would love to contribute in whatever way that I can and whatever it is that the users of this package require.

bitfield:
Great! Well, have a look at the open issues—there are quite a few of them, and most are waiting for some sort of design or further consideration, so I'm sure your input would be very helpful here!

PR #209, opened 2024-08-16
fix(race-conditon): fix data race condition when accessing `stderr` on pipe

### LINKED ISSUE #193: Race condition in Pipe.Exec
```
WARNING: DATA RACE
Read at 0x00c000036110 by goroutine 55:
  go/pkg/mod/github.com/bitfield/script@v0.22.0/script.go:388 +0x1e9
  github.com/bitfield/script.(*Pipe).Filter.func1()
      go/pkg/mod/github.com/bitfield/script@v0.22.0/script.go:484 +0xc7

Previous write at 0x00c000036110 by goroutine 49:
  github.com/bitfield/script.(*Pipe).WithStderr()
      go/pkg/mod/github.com/bitfield/script@v0.22.0/script.go:887 +0x798
```

There seems to be a data race with stdErr when executing the following piece of code in parallel (test with errgroup).
```go
stdOut := new(bytes.Buffer)
stdErr := new(bytes.Buffer)

if _, err := script.Exec(cmdLine).WithStdout(stdOut).WithStderr(stdErr).Stdout(); err != nil {
	return nil, err
}
```

mahadzaryab1:
There was a race condition that existed in `Pipe.Exec`, which spins up a goroutine that access `Pipe.stderr` without a lock. This PR fixes that by protecting `Pipe.stderr` with `Pipe.mu`. This PR fixes #193. 

**Testing**
Created a simple unit test as follows: 
```go
func TestDataRaceStdErr(t *testing.T) {
	stdOut := new(bytes.Buffer)
	stdErr := new(bytes.Buffer)

	_, err := script.Exec("echo").WithStdout(stdOut).WithStderr(stdErr).Stdout()
	if err != nil {
		t.Fatal(err)
	}
}
```
Running `go test -race` before the changes in this PR outputted the same warning seen in the issue. Running it again with the changes in the PR cause the tests to pass.

### DIFF
--- a/script.go
+++ b/script.go
@@ -387,7 +387,7 @@ func (p *Pipe) Exec(cmdLine string) *Pipe {
 		cmd.Stderr = w
 		pipeStderr := p.getStderr()
 		if pipeStderr != nil {
-			cmd.Stderr = p.stderr
+			cmd.Stderr = pipeStderr
 		}
 		err = cmd.Start()
 		if err != nil {
@@ -428,7 +428,7 @@ func (p *Pipe) ExecForEach(cmdLine string) *Pipe {
 			cmd.Stderr = w
 			pipeStderr := p.getStderr()
 			if pipeStderr != nil {
-				cmd.Stderr = p.stderr
+				cmd.Stderr = pipeStderr
 			}
 			err = cmd.Start()
 			if err != nil {
--- a/script.go
+++ b/script.go
@@ -27,13 +27,14 @@ import (
 // Pipe represents a pipe object with an associated [ReadAutoCloser].
 type Pipe struct {
 	// Reader is the underlying reader.
-	Reader         ReadAutoCloser
-	stdout, stderr io.Writer
-	httpClient     *http.Client
+	Reader     ReadAutoCloser
+	stdout     io.Writer
+	httpClient *http.Client
 
 	// because pipe stages are concurrent, protect 'err' and 'stderr'
-	mu  *sync.Mutex
-	err error
+	mu     *sync.Mutex
+	err    error
+	stderr io.Writer
 }
 
 // Args creates a pipe containing the program's command-line arguments from
--- a/script.go
+++ b/script.go
@@ -114,7 +114,7 @@ func Get(url string) *Pipe {
 
 // getStderr obtains the stderr writer on the pipe. This field
 // is protected by a mutex since stderr is accessed inside a
-// goroutine from [Pipe.Exec].
+// goroutine from [Pipe.Exec] and [Pipe.ExecForEach].
 func (p *Pipe) getStderr() io.Writer {
 	if p.mu == nil { // uninitialised pipe
 		return nil
@@ -808,7 +808,7 @@ func (p *Pipe) SetError(err error) {
 
 // setStderr sets the stderr writer on the pipe. This field
 // is protected by a mutex since stderr is accessed inside a
-// goroutine from [Pipe.Exec].
+// goroutine from [Pipe.Exec] and [Pipe.ExecForEach].
 func (p *Pipe) setStderr(stderr io.Writer) {
 	if p.mu == nil { // uninitialised pipe
 		return
--- a/script.go
+++ b/script.go
@@ -31,7 +31,7 @@ type Pipe struct {
 	stdout, stderr io.Writer
 	httpClient     *http.Client
 
-	// because pipe stages are concurrent, protect 'err'
+	// because pipe stages are concurrent, protect 'err' and 'stderr'
 	mu  *sync.Mutex
 	err error
 }
@@ -385,9 +385,12 @@ func (p *Pipe) Exec(cmdLine string) *Pipe {
 		cmd.Stdin = r
 		cmd.Stdout = w
 		cmd.Stderr = w
-		if p.stderr != nil {
+
+		pipeStderr := p.getStderr()
+		if pipeStderr != nil {
 			cmd.Stderr = p.stderr
 		}
+
 		err = cmd.Start()
 		if err != nil {
 			fmt.Fprintln(cmd.Stderr, err)
@@ -425,7 +428,8 @@ func (p *Pipe) ExecForEach(cmdLine string) *Pipe {
 			cmd := exec.Command(args[0], args[1:]...)
 			cmd.Stdout = w
 			cmd.Stderr = w
-			if p.stderr != nil {
+			pipeStderr := p.getStderr()
+			if pipeStderr != nil {
 				cmd.Stderr = p.stderr
 			}
 			err = cmd.Start()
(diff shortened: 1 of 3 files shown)


mahadzaryab1:
question: we currently do not run `go test` in the CI with the `-race` flag so I removed the regression test. Do we want to run the tests with the `-race` flag? If so, should I add the regression test back in?

bitfield:
If we can _have_ race conditions, we should definitely run CI tests with the race detector on!

mahadzaryab1:
> If we can _have_ race conditions, we should definitely run CI tests with the race detector on!

Added the `-race` flag to `ci.yml` and added a regression test.

script.go
```diff
@@ -385,7 +385,8 @@ func (p *Pipe) Exec(cmdLine string) *Pipe {
 		cmd.Stdin = r
 		cmd.Stdout = w
 		cmd.Stderr = w
-		if p.stderr != nil {
+		pipeStderr := p.getStderr()
+		if pipeStderr != nil {
 			cmd.Stderr = p.stderr
 		}
```
ccoVeille, line 391: I have a problem with the logic here

getStderr returns pipeStdErr


```suggestion
		pipeStderr := p.getStderr()
		if pipeStderr != nil {
 			cmd.Stderr = pipeStderr
 		}
```

While the code is somehow similar, even if there could remain some races, because we are accessing the p.stderr directly, no ?

script.go
```diff
@@ -425,7 +426,8 @@ func (p *Pipe) ExecForEach(cmdLine string) *Pipe {
 			cmd := exec.Command(args[0], args[1:]...)
 			cmd.Stdout = w
 			cmd.Stderr = w
-			if p.stderr != nil {
+			pipeStderr := p.getStderr()
+			if pipeStderr != nil {
 				cmd.Stderr = p.stderr
 			}
```
ccoVeille, line 432:
```suggestion
			pipeStderr := p.getStderr()
			if pipeStderr != nil {
 				cmd.Stderr = pipeStderr
 			}
```

script.go
ccoVeille: If not, the method and signature could be updated, because there is no need to return it, so,it's not a getter

ccoVeille (changes requested):

script.go
mahadzaryab1: @ccoVeille this was an oversight! thanks for catching - fixed now

script.go
```diff
@@ -31,7 +31,7 @@ type Pipe struct {
 	stdout, stderr io.Writer
 	httpClient     *http.Client
 
-	// because pipe stages are concurrent, protect 'err'
+	// because pipe stages are concurrent, protect 'err' and 'stderr'
```
bitfield, line 34: Notice how the `err` field is listed below the `mu` field, in its own 'block', separated from the other fields by a blank line. This is a nice way to show, lexically, that the mutex is 'over' these fields. To maintain this, let's move the `stderr` field in underneath the mutex too.

Is there any reason not to include the `stdout` field too? I mean, if a race condition exists on `stderr`, doesn't it also exist on `stdout`?

script.go
```diff
@@ -762,6 +764,30 @@ func (p *Pipe) SetError(err error) {
...
+}
+
+// getStderr obtains the stderr writer on the pipe. This field
+// is protected by a mutex since stderr is accessed inside a
+// goroutine from [Pipe.Exec].
+func (p *Pipe) getStderr() io.Writer {
+	if p.mu == nil { // uninitialised pipe
+		return nil
+	}
+	p.mu.Lock()
```
bitfield, line 786:
```suggestion
	p.mu.RLock()
```

I think we only need a read lock here.

script.go
mahadzaryab1: > Notice how the `err` field is listed below the `mu` field, in its own 'block', separated from the other fields by a blank line. This is a nice way to show, lexically, that the mutex is 'over' these fields. To maintain this, let's move the `stderr` field in underneath the mutex too.
addressed!
> Is there any reason not to include the `stdout` field too? I mean, if a race condition exists on `stderr`, doesn't it also exist on `stdout`?
The race condition exists on `stderr` because it is accessed inside a goroutine at https://github.com/bitfield/script/blob/master/script.go#L388-L390. The same is not true for `stdout` so I don't believe a race condition exists on `stdout`. We could still protect it with a mutex to match the behavior of `stdout` if we'd like however. Let me know what you think.

script.go
mahadzaryab1: @bitfield we can't do this since we're using a `sync.Mutex` instead of a `sync.RWMutex`

bitfield: This needs a little more thought, doesn't it? If we're just _retrieving_ the `stderr` writer from the `Pipe` struct, then to be sure we only need a read lock (and you've answered your own question, in a way: the solution would be to change the existing `sync.Mutex to a `sync.RWMutex`).

But the question arises whether it's safe for multiple concurrent `Exec`s to all write to the same writer. We can't know that it is. On the other hand, if we were to enforce serialization with a separate mutex for this error writer, then we'd lose the concurrency.

On balance I think we should let the user worry about that, don't you? After all, the configured stderr is overwhelmingly likely to be either `os.Stderr` or some file, in which case you might end up with interleaved writes looking messy, but no actual crashes.

As far as the `p.mu` goes, I think we can keep it as a write-lock for now: it doesn't seem like this is in the hot path of anything.

script_test.go
```diff
@@ -1850,6 +1850,19 @@ func TestReadReturnsErrorGivenReadErrorOnPipe(t *testing.T) {
 	}
 }
 
+// TestWithStdErrAfterExec is a regression test that was added to test against
+// a race condition for Pipe.stderr.
+func TestWithStdErrAfterExec(t *testing.T) {
```
bitfield, line 1855: Testing for data races is not an exact science, but it's generally a good idea to put the doubtful code under as much stress as possible—calling it a few hundred times from two or more concurrent goroutines, for example. I don't think you mentioned whether this test failed without your fix, with or without the race detector? I wouldn't have thought it would.

Can you put together a test that fails _fairly_ reliably without the mutex lock? I don't know if you've read my book [Know Go](https://bitfieldconsulting.com/books/generics) (few people have, so I don't blame you), but there's a section in there on "smoke testing" code for concurrency safety. Here's one of the examples:

```go
s := NewSetC(1, 2, 3)
var wg sync.WaitGroup
wg.Add(1)
go func() {
	for i := range 1000 {
		s.Add(i)
	}
	wg.Done()
}()
for i := range 1000 {
	_ = s.All()
}
wg.Wait()
fmt.Println("We made it!")
// Output:
// We made it!
```

bitfield (changes requested):
Looking good so far!

script.go
mahadzaryab1: Sounds good! Thank you for the detailed explanation @bitfield

script_test.go
mahadzaryab1: @bitfield Thanks for the detailed comment. Do we want a test that fails without the race detector on? The current test I've written fails with the race detector on but not without.

bitfield: If it fails with the race detector on, that's good enough!

bitfield:
@mahadzaryab1 I'm happy if you're happy—do you want to go ahead and do a final quality pass, checking documentation comments, alphabetical order, `gotestdox`-friendly test names, and so on? You know the house style now 😄

mahadzaryab1:
> @mahadzaryab1 I'm happy if you're happy—do you want to go ahead and do a final quality pass, checking documentation comments, alphabetical order, `gotestdox`-friendly test names, and so on? You know the house style now 😄

@bitfield Done! Let me know if there's anything I missed.

bitfield (approved):
Super nice, @mahadzaryab1! Thanks a lot. I'll wait till you're done on #208, and then we can roll all these changes up into a new release.

PR #210, opened 2024-08-16
feat(filters): add new filters for encoding and decoding base64 strings

### LINKED ISSUE #181: Add Base64 encode and decode methods

mahadzaryab1:
Added two new filters: `EncodeBase64` and `DecodeBase64` for encoding and decoding base64 strings. This PR closes #181.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ If you're already familiar with shell scripting and the Unix toolset, here is a 
 | `>`                | [`WriteFile`](https://pkg.go.dev/github.com/bitfield/script#Pipe.WriteFile) |
 | `>>`               | [`AppendFile`](https://pkg.go.dev/github.com/bitfield/script#Pipe.AppendFile) |
 | `$*`               | [`Args`](https://pkg.go.dev/github.com/bitfield/script#Args) |
+| `base64`           | [`DecodeBase64`](https://pkg.go.dev/github.com/bitfield/script#Pipe.DecodeBase64) / [`EncodeBase64`](https://pkg.go.dev/github.com/bitfield/script#Pipe.EncodeBase64) |
 | `basename`         | [`Basename`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Basename) |
 | `cat`              | [`File`](https://pkg.go.dev/github.com/bitfield/script#File) / [`Concat`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Concat) |
 | `curl`             | [`Do`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Do) / [`Get`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Get) / [`Post`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Post) |
@@ -290,7 +291,7 @@ Filters are methods on an existing pipe that also return a pipe, allowing you to
 | [`Basename`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Basename) | removes leading path components from each line, leaving only the filename |
 | [`Column`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Column) | Nth column of input |
 | [`Concat`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Concat) | contents of multiple files |
-| [`DecodeBase64`](https://pkg.go.dev/github.com/bitfield/script#Pipe.DecodeBase64) | decode base64 encoded input |
+| [`DecodeBase64`](https://pkg.go.dev/github.com/bitfield/script#Pipe.DecodeBase64) | input decoded from base64 |
 | [`Dirname`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Dirname) | removes filename from each line, leaving only leading path components |
 | [`Do`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Do) | response to supplied HTTP request |
 | [`EncodeBase64`](https://pkg.go.dev/github.com/bitfield/script#Pipe.EncodeBase64) | input encoded to base64 |
@@ -339,6 +340,7 @@ Sinks are methods that return some data from a pipe, ending the pipeline and ext
 
 | Version | New |
 | ----------- | ------- |
+| _next_  | [`DecodeBase64`](https://pkg.go.dev/github.com/bitfield/script#Pipe.DecodeBase64) / [`EncodeBase64`](https://pkg.go.dev/github.com/bitfield/script#Pipe.EncodeBase64) |
 | v0.22.0 | [`Tee`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Tee), [`WithStderr`](https://pkg.go.dev/github.com/bitfield/script#Pipe.WithStderr) |
 | v0.21.0 | HTTP support: [`Do`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Do), [`Get`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Get), [`Post`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Post) |
 | v0.20.0 | [`JQ`](https://pkg.go.dev/github.com/bitfield/script#Pipe.JQ) |
--- a/README.md
+++ b/README.md
@@ -313,6 +313,8 @@ Filters are methods on an existing pipe that also return a pipe, allowing you to
 | [`ReplaceRegexp`](https://pkg.go.dev/github.com/bitfield/script#Pipe.ReplaceRegexp) | matching text replaced with given string |
 | [`SHA256Sums`](https://pkg.go.dev/github.com/bitfield/script#Pipe.SHA256Sums) | SHA-256 hashes of each listed file |
 | [`Tee`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Tee) | input copied to supplied writers |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -313,6 +313,8 @@ Filters are methods on an existing pipe that also return a pipe, allowing you to
 | [`ReplaceRegexp`](https://pkg.go.dev/github.com/bitfield/script#Pipe.ReplaceRegexp) | matching text replaced with given string |
 | [`SHA256Sums`](https://pkg.go.dev/github.com/bitfield/script#Pipe.SHA256Sums) | SHA-256 hashes of each listed file |
 | [`Tee`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Tee) | input copied to supplied writers |
+| [`EncodeBase64`](https://pkg.go.dev/github.com/bitfield/script#Pipe.EncodeBase64) | input encoded to base64 |
```
bitfield, line 316: Could we keep these in alphabetical order if possible?

script.go
```diff
@@ -967,3 +968,36 @@ func newScanner(r io.Reader) *bufio.Scanner {
 	scanner.Buffer(make([]byte, 4096), math.MaxInt)
 	return scanner
 }
+
+// EncodeBase64 produces the base64 encoding of the input.
```
bitfield, line 972: Can we keep pipe methods all together, and preserve the existing alphabetical order? It just makes it easier to find stuff.

script.go
```diff
@@ -967,3 +968,36 @@ func newScanner(r io.Reader) *bufio.Scanner {
...
+func (p *Pipe) EncodeBase64() *Pipe {
+	return p.Filter(func(r io.Reader, w io.Writer) error {
+		output := new(strings.Builder)
```
bitfield, line 975: Do we actually need to go via a string?

script_test.go
```diff
@@ -1850,6 +1850,82 @@ func TestReadReturnsErrorGivenReadErrorOnPipe(t *testing.T) {
...
+			s:    "",
+			want: "\n",
+		},
+		{
+			name: "single line string",
+			s:    "hello world",
+			want: "aGVsbG8gd29ybGQ=\n",
+		},
+		{
+			name: "multi line string",
```
bitfield, line 1871: What do you think about adding some bytes cases? After all, it seems like it's exactly non-string data that people would want to encode to `base64`.

script_test.go
```diff
@@ -1850,6 +1850,82 @@ func TestReadReturnsErrorGivenReadErrorOnPipe(t *testing.T) {
...
+			s:    "",
+			want: "\n",
+		},
+		{
+			name: "single line output",
+			s:    "aGVsbG8gd29ybGQ=",
+			want: "hello world\n",
+		},
+		{
+			name: "multi line output",
```
bitfield, line 1909: Shall we use the same set of cases for testing both `Encode` and `Decode`? Also, what about a round-trip test?

bitfield (changes requested):
Great work!

script.go
mahadzaryab1: @bitfield changed this to use a `bytes.Buffer` - does this work better?

script_test.go
mahadzaryab1: @bitfield would you be able to clarify? do you want one test for both methods? and yes, i can add a round trip test!

script_test.go
mahadzaryab1: @bitfield what would these tests look like? do you want something like `str([]byte("hello world"))`?

script.go
bitfield: My point is really that `base64` has a byte-stream-based API (`Encoder`), and `Filter` is inherently a byte-stream-oriented function. Does it make sense to create an `Encoder` on `w` and `io.Copy` to it?

https://pkg.go.dev/encoding/base64#example-NewEncoder

script_test.go
bitfield: What I mean is, suppose you had some arbitrary `byte` data (not necessarily a UTF-8-encoded string), and you ran it through a pipe with `EncodeBase64`. That should produce more or less the same string as piping the same bytes through a tool like `base64`, shouldn't it?

script_test.go
bitfield: We don't have to combine the two tests into one; instead, we could move the test cases outside the function, so that both tests can share it.

```go
var base64Cases = ...

func TestDecodeBase64...
```

bitfield (changes requested):

script.go
mahadzaryab1: that's a very elegant solution! thanks for the suggestion

script_test.go
mahadzaryab1: @bitfield got it! i added a round trip test with a byte reader - let me know if this along the lines of what you were thinking

README.md
```diff
@@ -298,6 +298,8 @@ Filters are methods on an existing pipe that also return a pipe, allowing you to
 | [`Filter`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Filter) | user-supplied function filtering a reader to a writer |
 | [`FilterLine`](https://pkg.go.dev/github.com/bitfield/script#Pipe.FilterLine) | user-supplied function filtering each line to a string|
 | [`FilterScan`](https://pkg.go.dev/github.com/bitfield/script#Pipe.FilterScan) | user-supplied function filtering each line to a writer |
+| [`DecodeBase64`](https://pkg.go.dev/github.com/bitfield/script#Pipe.DecodeBase64) | decode base64 encoded input |
```
bitfield, line 301: Can we keep these in alphabetical order as well? It's an easy thing to overlook, but a list that's _mostly_ in order is very confusing for readers.

script.go
```diff
@@ -275,6 +276,20 @@ func (p *Pipe) CountLines() (lines int, err error) {
 	return lines, p.Error()
 }
 
+// DecodeBase64 produces the string represented by the base64 encoded input.
+func (p *Pipe) DecodeBase64() *Pipe {
+	return p.Filter(func(r io.Reader, w io.Writer) error {
+		decoder := base64.NewDecoder(base64.StdEncoding, r)
```
bitfield, line 282: The use of `StdEncoding` reminds me that there are other encodings; not that common, perhaps, but they exist. It would be a shame if some user wanted to use `URLEncoding`, for example, but couldn't. What if there were a way to provide the encoding you want if it _isn't_ `StdEncoding`?

script_test.go
```diff
@@ -1850,6 +1850,104 @@ func TestReadReturnsErrorGivenReadErrorOnPipe(t *testing.T) {
...
+		})
+	}
+}
+
+func TestEncodeAndDecodeBase64Bytes(t *testing.T) {
+	t.Parallel()
+	want := []byte{8, 0, 0, 16}
+	input := bytes.NewReader(want)
+
+	got, err := script.NewPipe().WithReader(input).EncodeBase64().DecodeBase64().Bytes()
```
bitfield, line 1941: We already have some round-trip tests (great!), so what about testing the expected encoded string produced by your byte slice? Use some out-of-band method to produce this (the `base64` tool, for example).

script_test.go
```diff
@@ -2284,6 +2382,20 @@ func ExampleSlice() {
...
 }
 
+func ExamplePipe_EncodeBase64() {
+	script.Echo("hello\nthere\nworld\n").EncodeBase64().Stdout()
+	// Output:
+	// aGVsbG8KdGhlcmUKd29ybGQK
+}
+
+func ExamplePipe_DecodeBase64() {
+	script.Echo("aGVsbG8KdGhlcmUKd29ybGQK").DecodeBase64().Stdout()
```
bitfield, line 2392: Just out of curiosity, how do we handle base64 input that's padded with one or two `=` characters?  `StdEncoding` does use padding, it seems, but we don't have any examples in the test cases. It would be weird if it didn't work, since we're using the `stdlib` for the implementation, of course, but that's the point of tests: they should specify behaviour without relying on any assumptions about implementation details.

bitfield (changes requested):

script_test.go
mahadzaryab1: @bitfield the `single line string` test case does have a decoded input with padding (`aGVsbG8gd29ybGQ=`). does this work?

script.go
mahadzaryab1: @bitfield I believe we have a couple of options here: 
- Take a `*base64.Encoding` as input to the method. I believe this isn't preferred based on the conversation in the issue but let me know if we want to revisit that. 
- We can create a `WithBase64Encoding` method that can override which encoding is used

Happy to implement either approach! Let me know what you think.

bitfield: We _could_ use the trick of taking `...*base64.Encoding`, so that you don't have to pass an argument if you don't want to. I always think this is a bit too clever, though, and it has some puzzling edge cases (what if you pass more than one argument?)

Second-guessing myself again, I can't really imagine anyone wanting to use anything other than `StdEncoding`. Let's stick with that. If we _wanted_ to add an "optional" encoding argument later, it wouldn't break existing code (unless it referred to `p.EncodeBase64` as a function value, of course, but you can't win them all).

bitfield (approved):
This looks great! A very neat job. Are you happy for me to pick this up and make the usual minor tweaks, commas and newlines, and final polish? I plan to roll this up along with your other excellent PRs into the next release.

mahadzaryab1:
@bitfield Thank you so much! And yes absolutely, feel free to pick it up and make any tweaks that are needed.

PR #211, opened 2024-08-18
refactor: return error on pipe from `Wait` method after completion of pipe

### LINKED ISSUE #123: ExecForEach().Error() doesn't execute
Hello, I'm currently trying to migrate from some `build.sh` and `build.ps1` files to a [magefile](https://magefile.org/) using `script`.

Maybe I misunderstood the purpose of `Error()`, but when I don't need the output of a command (as `String()` would return), I thought `Error()` would basically ignore the output, but still return the error. But it seems that when chained after an `ExecForEach`, the commands in that are not executed.

For example, let's say we have files `coverage.txt` and `foo/coverage.txt` and I want to remove them.
Then the following works:

```go
	_, err := script.FindFiles(".").Match("coverage.txt").ExecForEach("rm ./{{.}}").String()
	return err
```

But the following doesn't:

```go
	return script.FindFiles(".").Match("coverage.txt").ExecForEach("rm ./{{.}}").Error()
```

Is that expected?

mahadzaryab1:
Updated the `Wait` method to return the error present on the pipe after completion. This PR closes #123.

### DIFF
--- a/script.go
+++ b/script.go
@@ -858,6 +858,14 @@ func (p *Pipe) Wait() {
 	}
 }
 
+// WaitError reads the pipe to completion and returns any error present on
+// the pipe, or nil otherwise. This is mostly useful for waiting until
+// concurrent filters have completed (see [Pipe.Filter]).
+func (p *Pipe) WaitError() error {
+	p.Wait()
+	return p.Error()
+}
+
 // WithError sets the error err on the pipe.
 func (p *Pipe) WithError(err error) *Pipe {
 	p.SetError(err)
--- a/script.go
+++ b/script.go
@@ -859,7 +859,6 @@ func (p *Pipe) Wait() error {
 	if err != nil {
 		p.SetError(err)
 	}
-
 	return p.Error()
 }
 
--- a/script.go
+++ b/script.go
@@ -348,6 +348,9 @@ func (p *Pipe) Echo(s string) *Pipe {
 }
 
 // Error returns any error present on the pipe, or nil otherwise.
+// Error is not a sink and does not wait until the pipe reaches
+// completion. To wait for completion before returning the error,
+// see [Pipe.WaitError].
 func (p *Pipe) Error() error {
 	if p.mu == nil { // uninitialised pipe
 		return nil
--- a/script.go
+++ b/script.go
@@ -350,7 +350,7 @@ func (p *Pipe) Echo(s string) *Pipe {
 // Error returns any error present on the pipe, or nil otherwise.
 // Error is not a sink and does not wait until the pipe reaches
 // completion. To wait for completion before returning the error,
-// see [Pipe.WaitError].
+// see [Pipe.Wait].
 func (p *Pipe) Error() error {
 	if p.mu == nil { // uninitialised pipe
 		return nil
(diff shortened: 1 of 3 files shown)


script_test.go
```diff
@@ -1850,6 +1850,22 @@ func TestReadReturnsErrorGivenReadErrorOnPipe(t *testing.T) {
...
+	p := script.Echo("a\nb\nc\n").ExecForEach("{{invalid template syntax}}")
+	if p.Wait() == nil {
+		t.Error("want error, got nil")
+	}
+}
+
+func TestWaitForNoErrorOnPipe(t *testing.T) {
+	t.Parallel()
+	p := script.Echo("a\nb\nc\n").ExecForEach("echo \"{{.}}\"")
+	if err := p.Wait(); err != nil {
```
ccoVeille, line 1864: I would expect the test to not only check for the error, but also to validate the error type.

I'm talking about checking the error type with errors.Is

Sometimes a refactoring could change the behavior and an unexpected error will be returned. But then the tests won't see it, so users will face it later when merged

script.go
```diff
@@ -848,14 +851,16 @@ func (p *Pipe) Tee(writers ...io.Writer) *Pipe {
...
+// Wait reads the pipe to completion and returns any error present on
+// the pipe, or nil otherwise. This is mostly useful for waiting until
+// concurrent filters have completed (see [Pipe.Filter]).
+func (p *Pipe) Wait() error {
 	_, err := io.Copy(io.Discard, p)
 	if err != nil {
 		p.SetError(err)
 	}
+
+	return p.Error()
```
ccoVeille, line 863: Is there a difference with doing this?


```suggestion
 		p.SetError(err)
 		return err
 	}

	return nil
```

It would be more readable

mahadzaryab1: i think currently there would be no difference but its nice to have a standard place for fetching the error in case we ever add anything to that method

script_test.go
mahadzaryab1: I'm not entirely sure what value we get from checking the specific error type. The goal of this test is to just test that if there is an error present on the pipe, we can get it using `Wait`.

ccoVeille: It's true that you are replicating an existing test.

I would have expected that both are catching the expected error for an invalid template.

The fact, I wasn't able to convince you earlier, let me think either I cannot convince you, or that I'm not able to explain it.

I don't mind if you keep the current code, it's a test. It was also a side remark.

bitfield: > I would expect the test to not only check for the error, but also to validate the error type.

That's not a crazy idea, but it's not something I generally do unless returning some _specific_ error type is an important part of the function's behaviour. If, on the other hand, as is usually the case, all that matters is that it returns _some_ error, then checking it's not `nil` is good enough.

I think it's wise not to test more than you really need to, for fear of ending up with overly fragile tests.

script.go
bitfield: > Is there a difference with doing this?

But what if there's _already_ some error on the _pipe_, but the `io.Copy` was successful? In that case we would return `nil`, losing the existing `p.Error()`.

script.go
```diff
@@ -848,14 +851,16 @@ func (p *Pipe) Tee(writers ...io.Writer) *Pipe {
...
-func (p *Pipe) Wait() {
+// Wait reads the pipe to completion and returns any error present on
+// the pipe, or nil otherwise. This is mostly useful for waiting until
+// concurrent filters have completed (see [Pipe.Filter]).
+func (p *Pipe) Wait() error {
 	_, err := io.Copy(io.Discard, p)
 	if err != nil {
 		p.SetError(err)
 	}
+
```
bitfield, line 862: Just a tiny point of house style, but we don't use blank lines within functions (there's a lot of small functions in `script`, and blank lines scattered willy-nilly throughout them would be in danger of doubling the length of the code).

bitfield (changes requested):
I think this is about ready to go once you've rebased it on `master`!

script.go
ccoVeille: Oh good catch

script_test.go
ccoVeille: Indeed. I understand your point

script_test.go
```diff
@@ -1850,6 +1850,22 @@ func TestReadReturnsErrorGivenReadErrorOnPipe(t *testing.T) {
 	}
 }
 
+func TestWaitForErrorOnPipe(t *testing.T) {
```
bitfield, line 1853: If you try `gotestdox` here, as discussed on the other PR, you'll see why this test name doesn't quite fit with the others:

```
 ✔ Error returns error set by previous pipe stage (0.00s)
...

 ✔ Wait reads pipe source to completion (0.00s)
...
 ✔ Wait for error on pipe (0.00s)
 ✔ Wait for no error on pipe (0.00s)
```

See what I mean? It's not a sentence describing the effect that `Wait` _has_. (This is why I find `gotestdox` so useful as a thinking tool: it forces me to describe the behaviour of the thing I'm implementing.)

bitfield (changes requested):

script_test.go
mahadzaryab1: Thanks for the great suggestion @bitfield! changed the test names to conform to your suggestion. Let me know what you think.

bitfield (approved):
Great job! Thanks a lot.

PR #215, opened 2024-09-28
feat: implement `Hash` and `HashSums` to performing hashing based on user provided hashers

### LINKED ISSUE #113: Add SHA-512 Hashing Sink

mahadzaryab1:
- Added a `Hash` sink that returns the hex encoded hash of the pipe contents based on the `hash.Hash` argument passed into the method
- Added a `HashSums` filter that produces the hex encoded hash of the files in the path 

This PR resolves #113, #130.


script.go
```diff
@@ -650,6 +651,38 @@ func (p *Pipe) Get(url string) *Pipe {
...
+func (p *Pipe) Hash(hasher hash.Hash) (string, error) {
+	if p.Error() != nil {
+		return "", p.Error()
+	}
+	_, err := io.Copy(hasher, p)
+	if err != nil {
+		p.SetError(err)
+		return "", err
+	}
+	return hex.EncodeToString(hasher.Sum(nil)), p.Error()
```
bitfield, line 665: _Can_ `p.Error()` be non-nil at this point? If not, we can just return nil explicitly.

script_test.go
```diff
@@ -2276,6 +2373,22 @@ func ExamplePipe_Get() {
...
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(sum)
+	// Output:
+	// 309ecc489c12d6eb4cc40f50c902f2b4d0ed77ee511a7c7a9bcd3ca86d4cd86f989dd35bc5ff499670da34255b45b0cfd830e81f605dcf7dc5542e93ae9cd76f
+}
+
+func ExamplePipe_HashSums() {
+	script.Echo("testdata/test.txt").HashSums(sha512.New()).Stdout()
```
bitfield, line 2387: Would it be clearer if we used two files here, to show that the method produces one hash per file?

testdata/sha512Sum.input.txt
```diff
@@ -0,0 +1,2 @@
+hello
```
bitfield, line 1: We don't really need a separate file just for this, do we?

bitfield (changes requested):
Great! Could you also please [deprecate](https://go.dev/wiki/Deprecated) `SHA256Sum(s)` in favour of the new methods?

script_test.go
mahadzaryab1: done!

script.go
mahadzaryab1: done!

README.md
```diff
@@ -317,6 +317,7 @@ Filters are methods on an existing pipe that also return a pipe, allowing you to
 | [`First`](https://pkg.go.dev/github.com/bitfield/script#Pipe.First) | first N lines of input |
 | [`Freq`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Freq) | frequency count of unique input lines, most frequent first |
 | [`Get`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Get) | response to HTTP GET on supplied URL |
+| [`HashSums`](https://pkg.go.dev/github.com/bitfield/script#Pipe.HashSums) | hashes of each listed file |
```
bitfield, line 320: Shall we also remove SHA256Sum(s) from these lists, then?

script.go
```diff
@@ -650,6 +651,38 @@ func (p *Pipe) Get(url string) *Pipe {
 	return p.Do(req)
 }
 
+// Hash returns the hex-encoded hash of the entire contents of the
```
bitfield, line 654: Maybe add a cross-reference to HashSums here, and vice versa?

script.go
```diff
@@ -816,36 +849,19 @@ func (p *Pipe) SetError(err error) {
 
 // SHA256Sum returns the hex-encoded SHA-256 hash of the entire contents of the
 // pipe, or an error.
+// Deprecated: SHA256Sums has been deprecated by Hash. To get the SHA 256
```
bitfield, line 852: Can you make this a link to Hash, using square brackets?

script.go
```diff
@@ -816,36 +849,19 @@ func (p *Pipe) SetError(err error) {
...
-		return "", err
-	}
-	return hex.EncodeToString(hasher.Sum(nil)), p.Error()
+	return p.Hash(sha256.New())
 }
 
 // SHA256Sums reads paths from the pipe, one per line, and produces the
 // hex-encoded SHA-256 hash of each corresponding file, one per line. Any files
 // that cannot be opened or read will be ignored.
+// Deprecated: SHA256Sums has been deprecated by HashSums. To get the SHA 256
```
bitfield, line 861: Ditto

script_test.go
```diff
@@ -2013,6 +2016,100 @@ func TestWithStdErr_IsConcurrencySafeAfterExec(t *testing.T) {
...
+			want:   "8710339dcb6814d0d9d2290ef422285c9322b7163951f9a0ca8f883d3305286f44139aa374848e4174f5aada663027e4548637b6d19894aec4fb6c46a139fbf9",
+		},
+		{
+			name:   "for string containing newline with SHA 256 hasher",
+			input:  "The tao that can be told\nis not the eternal Tao",
+			hasher: sha256.New(),
+			want:   "788542cb92d37f67e187992bdb402fdfb68228a1802947f74c6576e04790a688",
+		},
+	}
+
```
bitfield, line 2050: Remove blank line

bitfield (changes requested):
Nice work!

script.go
```diff
@@ -816,36 +851,19 @@ func (p *Pipe) SetError(err error) {
 
 // SHA256Sum returns the hex-encoded SHA-256 hash of the entire contents of the
 // pipe, or an error.
+// Deprecated: SHA256Sum has been deprecated by [Pipe.Hash]. To get the SHA 256
```
bitfield, line 854:
```suggestion
// Deprecated: SHA256Sum has been deprecated by [Pipe.Hash]. To get the SHA-256
```

script.go
```diff
@@ -816,36 +851,19 @@ func (p *Pipe) SetError(err error) {
...
-		return "", err
-	}
-	return hex.EncodeToString(hasher.Sum(nil)), p.Error()
+	return p.Hash(sha256.New())
 }
 
 // SHA256Sums reads paths from the pipe, one per line, and produces the
 // hex-encoded SHA-256 hash of each corresponding file, one per line. Any files
 // that cannot be opened or read will be ignored.
+// Deprecated: SHA256Sums has been deprecated by [Pipe.HashSums]. To get the SHA 256
```
bitfield, line 863:
```suggestion
// Deprecated: SHA256Sums has been deprecated by [Pipe.HashSums]. To get the SHA-256
```

script.go
```diff
@@ -816,36 +851,19 @@ func (p *Pipe) SetError(err error) {
...
+// encoding for the paths in the pipe, call `HashSums(sha256.new())`
```
bitfield, line 864:
```suggestion
// hash for each file path in the pipe, call `HashSums(sha256.new())`
```

bitfield (changes requested):
Great! Just a couple of tweaks needed.

script.go
```diff
@@ -650,6 +651,40 @@ func (p *Pipe) Get(url string) *Pipe {
...
+	}
+	return hex.EncodeToString(hasher.Sum(nil)), nil
+}
+
+// HashSums reads paths from the pipe, one per line, and produces the
+// hex-encoded hash of each corresponding file based on the provided hasher,
+// one per line. Any files that cannot be opened or read will be ignored.
+// To perform hashing on the contents of the pipe, see [Pipe.Hash].
+func (p *Pipe) HashSums(hasher hash.Hash) *Pipe {
+	return p.FilterScan(func(line string, w io.Writer) {
```
bitfield, line 674: Sorry about the last-minute thought, but would it make sense to use `FilterLine' here instead? That's the simple case when there's one line in, one line out. `FilterScan` is more flexible, but you have to print to the writer yourself. I should really have used `FilterLine` to implement `SHA256Sums` in the first place; I'm not sure why I didn't.

bitfield: Maybe it's because `FilterLine` always produces a line, even if it's empty. What do you think makes more sense for files we can't hash? Print an empty line, so the output lines match up with the input, or just produce no output at all?

mahadzaryab1: `FilterLine` could work but I'm just wondering if we should make that change because of backwards compatibility purposes? This would be a breaking change for the existing functions - is that okay?

mahadzaryab1: @bitfield I made the change. We can always revert it if we decide we don't want to do it.

bitfield: That's a good point about changing the output. I think let's stick with `FilterScan`, then, and if it makes sense to change the behaviour in the future, we'll do it then. "Break one thing at a time" is a good motto!

mahadzaryab1: done

bitfield (approved):
Great job! Thanks again.

PR #216, opened 2024-10-19
fix(slice): fix bug in slice that produces new line for empty input slice

### LINKED ISSUE #110: `ListFiles()` Produces an Unwanted Newline Character When Zero Files Are Matched

mahadzaryab1:
This PR fixes #110 by handling an empty input into the `Slice` function.

Added a new unit test and consolidated it with the existing unit test. 
```
 ✔ Slice (0.00s)
 ✔ Slice given empty slice produces empty pipe (0.00s)
 ✔ Slice produces elements of specified slice one per line (0.00s)
```

### DIFF
--- a/script.go
+++ b/script.go
@@ -182,6 +182,9 @@ func Post(url string) *Pipe {
 
 // Slice creates a pipe containing each element of s, one per line.
 func Slice(s []string) *Pipe {
+	if len(s) == 0 {
+		return NewPipe()
+	}
 	return Echo(strings.Join(s, "\n") + "\n")
 }
 
--- a/script_test.go
+++ b/script_test.go
@@ -1375,15 +1375,34 @@ func TestReadAutoCloser_ReadsAllDataFromSourceAndClosesItAutomatically(t *testin
 	}
 }
 
-func TestSliceProducesElementsOfSpecifiedSliceOnePerLine(t *testing.T) {
+func TestSlice_(t *testing.T) {
 	t.Parallel()
-	want := "1\n2\n3\n"
-	got, err := script.Slice([]string{"1", "2", "3"}).String()
-	if err != nil {
-		t.Fatal(err)
+	tests := []struct {
+		name  string
+		input []string
+		want  string
+	}{
+		{
+			name:  "produces elements of specified slice one per line",
+			input: []string{"1", "2", "3"},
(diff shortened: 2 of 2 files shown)


script_test.go
```diff
@@ -1375,15 +1375,34 @@ func TestReadAutoCloser_ReadsAllDataFromSourceAndClosesItAutomatically(t *testin
...
 }
 
-func TestSliceProducesElementsOfSpecifiedSliceOnePerLine(t *testing.T) {
+func TestSlice_(t *testing.T) {
 	t.Parallel()
-	want := "1\n2\n3\n"
-	got, err := script.Slice([]string{"1", "2", "3"}).String()
-	if err != nil {
-		t.Fatal(err)
+	tests := []struct {
```
bitfield, line 1380: It seems a shame to have a table test with just two cases, doesn't it? Especially as they're *different* cases, not two examples of the same behaviour.

Wouldn't you rather just structure this as two separate tests? That tends to make each test clearer, more focused, and easier to understand.

bitfield (changes requested):
Great job! Not to minimise your work in the least, which is very welcome, but it's a little surprising that a fix like this has taken this long, isn't it?

script_test.go
mahadzaryab1: done!

script.go
```diff
@@ -182,6 +182,9 @@ func Post(url string) *Pipe {
 
 // Slice creates a pipe containing each element of s, one per line.
```
bitfield, line 183:
```suggestion
// Slice creates a pipe containing each element of s, one per line. If s is
// empty or nil, then the pipe is empty.
```

bitfield (approved):
Awesome! Thanks a lot!

PR #217, opened 2024-11-02
[fix] change behaviour of `FindFiles` to ignore errors 

### LINKED ISSUE #99: FindFiles exits and returns no files when it can't open a directory

mahadzaryab1:
- Replaced `filepath.Walk` with the newer `fs.WalkDir`
- Changed the behaviour of `FindFiles` to only return an error if no paths were found
- Fixes #99

### DIFF
--- a/script.go
+++ b/script.go
@@ -95,6 +95,9 @@ func FindFiles(dir string) *Pipe {
 	var paths []string
 	err := fs.WalkDir(os.DirFS(dir), ".", func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
+			if os.IsPermission(err) {
+				return fs.SkipDir
+			}
 			return err
 		}
 		if !d.IsDir() {
--- a/script.go
+++ b/script.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"hash"
 	"io"
+	"io/fs"
 	"math"
 	"net/http"
 	"os"
@@ -92,12 +93,12 @@ func File(path string) *Pipe {
 //	test/2.txt
 func FindFiles(dir string) *Pipe {
 	var paths []string
-	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+	err := fs.WalkDir(os.DirFS(dir), ".", func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
 		}
-		if !info.IsDir() {
-			paths = append(paths, path)
+		if !d.IsDir() {
+			paths = append(paths, filepath.Join(dir, path))
 		}
 		return nil
 	})
--- a/script.go
+++ b/script.go
@@ -93,20 +93,19 @@ func File(path string) *Pipe {
 //	test/2.txt
 func FindFiles(dir string) *Pipe {
 	var paths []string
-	err := fs.WalkDir(os.DirFS(dir), ".", func(path string, d fs.DirEntry, err error) error {
+	var innerErr error
+	fs.WalkDir(os.DirFS(dir), ".", func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
-			if os.IsPermission(err) {
-				return fs.SkipDir
-			}
-			return err
+			innerErr = err
+			return fs.SkipDir
 		}
 		if !d.IsDir() {
 			paths = append(paths, filepath.Join(dir, path))
 		}
 		return nil
 	})
-	if err != nil && len(paths) == 0 {
-		return NewPipe().WithError(err)
+	if innerErr != nil && len(paths) == 0 {
+		return NewPipe().WithError(innerErr)
 	}
 	return Slice(paths)
 }
--- a/script.go
+++ b/script.go
@@ -77,8 +77,12 @@ func File(path string) *Pipe {
 }
 
 // FindFiles creates a pipe listing all the files in the directory dir and its
-// subdirectories recursively, one per line, like Unix find(1). If dir doesn't
-// exist or can't be read, the pipe's error status will be set.
+// subdirectories recursively, one per line, like Unix find(1).
+// If an error occurs while reading the directory, the behaviour is as follows:
+//   - If no files have been found (the directory is empty or inaccessible), the
+//     resulting Pipe will have its error status set to the encountered error.
+//   - If at least one file is found, the error is ignored, and the resulting Pipe will
+//     contain the successfully found file paths.
 //
 // Each line of the output consists of a slash-separated path, starting with
 // the initial directory. For example, if the directory looks like this:
--- a/script.go
+++ b/script.go
@@ -102,7 +102,7 @@ func FindFiles(dir string) *Pipe {
 		}
 		return nil
 	})
-	if err != nil {
+	if err != nil && len(paths) == 0 {
 		return NewPipe().WithError(err)
 	}
 	return Slice(paths)
--- a/script.go
+++ b/script.go
@@ -78,11 +78,8 @@ func File(path string) *Pipe {
 
 // FindFiles creates a pipe listing all the files in the directory dir and its
 // subdirectories recursively, one per line, like Unix find(1).
-// If an error occurs while reading the directory, the behaviour is as follows:
-//   - If no files have been found (the directory is empty or inaccessible), the
-//     resulting Pipe will have its error status set to the encountered error.
-//   - If at least one file is found, the error is ignored, and the resulting Pipe will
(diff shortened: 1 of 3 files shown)


mahadzaryab1:
@bitfield Do you have any thoughts on how we should test the case where there is an error but we get more than one path?

bitfield:
That's a tricky one! What about creating a tempdir and then setting up files `a`, `b`, and `c` within it, of which the permissions of `c` are `0o000`? Would that work?

mahadzaryab1:
> That's a tricky one! What about creating a tempdir and then setting up files `a`, `b`, and `c` within it, of which the permissions of `c` are `0o000`? Would that work?

@bitfield I gave it a shot but I can't seem to produce that case. Changing the permissions of the file doesn't prevent the path from being listed. Let me know if you have any thoughts on how to proceed.

bitfield:
Remember `fs.WalkDir` only calls your func for directories. It doesn't try to read files, so you won't ever receive a non-nil error for permissions on a _file_, only on a directory.

mahadzaryab1:
@bitfield added a test case - can you take another look?

script_test.go
```diff
@@ -1286,6 +1286,39 @@ func TestFindFiles_InNonexistentPathReturnsError(t *testing.T) {
...
+	}
+	restrictedDirPath := filepath.Join(tmpDir, "restricted_dir")
+	if err := os.Mkdir(restrictedDirPath, os.ModePerm); err != nil {
+		t.Fatal(err)
+	}
+	fileBPath := filepath.Join(restrictedDirPath, "file_b.txt")
+	if err := os.WriteFile(fileBPath, []byte("hello again!"), os.ModePerm); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.Chmod(restrictedDirPath, 0o000); err != nil {
```
mahadzaryab1, line 1304: @bitfield this test fails on windows since chmod is a unix command. thoughts on how to proceed? should we make this a unix-only test?

bitfield: Yes, Windows doesn't have the concept of file permissions as Go understands them—this is Unix-only.

script_test.go
```diff
@@ -1286,6 +1286,39 @@ func TestFindFiles_InNonexistentPathReturnsError(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { os.Chmod(restrictedDirPath, os.ModePerm) })
```
bitfield, line 1307: This tempdir will be deleted anyway.

script_test.go
```diff
@@ -1286,6 +1286,39 @@ func TestFindFiles_InNonexistentPathReturnsError(t *testing.T) {
...
+	p := script.FindFiles(tmpDir)
```
bitfield, line 1308:
```suggestion
	got, err := script.FindFiles(tmpDir).String()
```

script_test.go
mahadzaryab1: @bitfield I need to do this so that this dir can get cleaned up. Otherwise, the cleanup function throws a permissions error. Let me know if you'd like to do this a different way. 
```
TempDir RemoveAll cleanup: openfdat /var/folders/mn/31m4g_6j7qz26m7_00mgsbrc0000gn/T/TestFindFiles_DoesNotErrorWhenSubDirectoryIsNotReadable1455531535/001/restricted_dir: permission denied
```

bitfield: That's because you're also creating a file _inside_ the restricted directory. So `RemoveAll` isn't able to delete the file, because that requires modifying the directory, which you don't have permission to do.

You don't need this file for the test (and `FindFIles` won't be able to even know it exists, because of the directory permissions). Without the file, you don't need the `Cleanup` func, because the automatic `RemoveAll` succeeds.

bitfield:
This test covers the case where the induced error is on the last path that `FindFiles` visits. Unfortunately, this just happens to miss a logic bug in the implementation of `FindFiles`. If you change the test (for example, renaming `restricted_dir` to `a_restricted_dir`, so that it's visited before `file_a`), you'll see what I mean.

script_test.go
mahadzaryab1: @bitfield If I don't add the file - then wouldn't the test pass regardless because FindFiles doesn't print directories. I was trying to show that there's a path inside the restricted dir that is not printed.

mahadzaryab1:
> This test covers the case where the induced error is on the last path that `FindFiles` visits. Unfortunately, this just happens to miss a logic bug in the implementation of `FindFiles`. If you change the test (for example, renaming `restricted_dir` to `a_restricted_dir`, so that it's visited before `file_a`), you'll see what I mean.

Ah okay I see. I added a check in `FindFiles` to skip the directory if its a permission error. What do you think?

script_test.go
bitfield: Well, according to your test name, the _test_ is about whether or not `FindFiles` returns an error if it encounters one on its walk. If the error happens to be "can't list files in directory", then we hardly need to test that `FindFIles` doesn't return those files: how could it? Testing the impossible doesn't catch bugs (testing _assumptions_, on the other hand, is an excellent idea and we should always do it).

bitfield:
Well, I don't think it matters what the error is. The point is that if we encounter _any_ errors during the walk, we should only report them if we haven't also found some valid files.

mahadzaryab1:
> Well, I don't think it matters what the error is. The point is that if we encounter _any_ errors during the walk, we should only report them if we haven't also found some valid files.

I see - should we then always return `fs.SkipDir` when we encounter an error during the walk?

script_test.go
mahadzaryab1: @bitfield got it - thanks for the explanation! i removed the file for the test

bitfield:
> should we then always return `fs.SkipDir` when we encounter an error during the walk?

Well, there are only four possibilities:

1. Return `SkipAll` - that's no good, because it stops walking altogether
2. Return the error - ditto
3. Return `nil` - no good, because it loses the error
4. Return `SkipDir`

So I don't think we actually have a choice to make here, unless there's another possibility you can think of?

mahadzaryab1:
> > should we then always return `fs.SkipDir` when we encounter an error during the walk?
> 
> Well, there are only four possibilities:
> 
> 1. Return `SkipAll` - that's no good, because it stops walking altogether
> 2. Return the error - ditto
> 3. Return `nil` - no good, because it loses the error
> 4. Return `SkipDir`
> 
> So I don't think we actually have a choice to make here, unless there's another possibility you can think of?

@bitfield Good point. But if we always run `SkipDir`, then we lose the error as well because WalkDir will swallow the error.

bitfield:
Right. But the `WalkFunc` can be a closure, so we can save the error in some variable in the outer scope before returning `SkipDir`.

mahadzaryab1:
> Right. But the `WalkFunc` can be a closure, so we can save the error in some variable in the outer scope before returning `SkipDir`.

@bitfield Addressed. Let me know what you think.

bitfield:
Great! This code looks solid. Would you like to update the documentation a little to explain this (sensible, but non-obvious) error behaviour?

mahadzaryab1:
> Great! This code looks solid. Would you like to update the documentation a little to explain this (sensible, but non-obvious) error behaviour?

Done!

script.go
```diff
@@ -76,8 +77,12 @@ func File(path string) *Pipe {
 }
 
 // FindFiles creates a pipe listing all the files in the directory dir and its
-// subdirectories recursively, one per line, like Unix find(1). If dir doesn't
-// exist or can't be read, the pipe's error status will be set.
+// subdirectories recursively, one per line, like Unix find(1).
+// If an error occurs while reading the directory, the behaviour is as follows:
```
bitfield, line 81: This is good, but I think we can be a bit more concise. What about:

```
// Errors are ignored unless no files are found (in which case the pipe's error
// status will be set to the last error encountered). 
```

mahadzaryab1: @bitfield done!

bitfield:
Super nice! Thanks for steering this one, great work.

PR #224, opened 2025-02-26
Implemented `MatchRegexpWithLineNumber()` and `MatchWithLineNumber()`

jessp01:
These behave as `grep -n` does: output the matched line prefixed by the line number.

**Simple test:**

> /tmp/test.txt
```
Sample data:
This is a test of MatchRegexpWithLineNumber() and MatchWithLineNumber()
They behave as the `grep -n` would
MatchRegexpWithLineNumber() will also find this TEST occurrence because it's case-insensitive
```

code:
```go
func main() {
        filePath := "/tmp/test.txt"
        pattern := "test"
        str, _ := script.File(filePath).MatchWithLineNumber(pattern).First(3).String()
        fmt.Printf("Testing MatchWithLineNumber with %s...\noutput is:\n%s\n", pattern, str)
        r := regexp.MustCompile("(?i)" + regexp.QuoteMeta(pattern))
        str, _ = script.File(filePath).MatchRegexpWithLineNumber(r).String()
        fmt.Printf("Testing MatchRegexpWithLineNumber with %s...\noutput is:\n%s\n", pattern, str)
}

```

Output:
```
Testing MatchWithLineNumber with test...
output is:
2:This is a test of MatchRegexpWithLineNumber() and MatchWithLineNumber()

Testing MatchRegexpWithLineNumber with test...
output is:
2:This is a test of MatchRegexpWithLineNumber() and MatchWithLineNumber()
4:MatchRegexpWithLineNumber() will also find this TEST occurrence because it's case-insensitive
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ Filters are methods on an existing pipe that also return a pipe, allowing you to
 | [`SHA256Sums`](https://pkg.go.dev/github.com/bitfield/script#Pipe.SHA256Sums) | SHA-256 hashes of each listed file |
 | [`Tee`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Tee) | input copied to supplied writers |
 
-Note that filters run concurrently, rather than producing nothing until each stage has fully read its input. This is convenient for executing long-running comands, for example. If you do need to wait for the pipeline to complete, call [`Wait`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Wait).
+Note that filters run concurrently, rather than producing nothing until each stage has fully read its input. This is convenient for executing long-running commands, for example. If you do need to wait for the pipeline to complete, call [`Wait`](https://pkg.go.dev/github.com/bitfield/script#Pipe.Wait).
 
 ## Sinks
 
--- a/script.go
+++ b/script.go
@@ -512,6 +512,20 @@ func (p *Pipe) FilterScan(filter func(string, io.Writer)) *Pipe {
 	})
 }
 
+// FilterScanTrackLine behaves the same way as FilterScan() but also returns the
+// line number of the processed input
+func (p *Pipe) FilterScanTrackLine(filter func(string, io.Writer, int)) *Pipe {
+	return p.Filter(func(r io.Reader, w io.Writer) error {
+		scanner := newScanner(r)
+		line := 1
+		for scanner.Scan() {
+			filter(scanner.Text(), w, line)
+			line++
+		}
+		return scanner.Err()
+	})
+}
+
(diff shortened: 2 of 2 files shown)


bitfield:
Hey @jessp01, thanks for the PR!

This is a neat idea. Could you post a comment with some example code showing how this feature would solve a user's problem? I can't think of any situation in which I've needed to use the equivalent of `grep -n` recently, but I'm sure there are some.

jessp01:
Hi @bitfield ,

Yes, as it so happens, I am working on a [glow feature](https://github.com/charmbracelet/glow/pull/705) I need this for right now:)

You can see a [screencast](https://asciinema.org/a/705359) where I demonstrate it.

I'm using your module [here](https://github.com/jessp01/glow/blob/README/ui/stash.go#L894)
The new `MatchRegexpWithLineNumber()` will enable me to open the relevant file in the line where the first match was found.

Later, I will also use it to introduce search functionality in the built-in glow pager (when it displays a given file).

PS
Had Go supported optional params or method overloading, I could have revised the original Match methods without breaking the API but since it doesn't, I had to introduce these two new methods.

bitfield:
What I'm really looking for is an example of code that uses `MatchRegexpWithLineNumber` (for example). Then we can see what it would look like in context and what kind of problems it solves.

jessp01:
Well, as I wrote, I'd replace the call to `MatchRegexp()` here: https://github.com/jessp01/glow/blob/README/ui/stash.go#L894 with `MatchRegexpWithLineNumber()`, split the output by ":" so I have the line number in which the string appears and the actual string and use the former to open the file and move to the line number in question.

And actually, as I write this, I think it would make sense to go a step further and revise these new methods so that they return the string and line number as separate values. If you agree, I'll make the change.

bitfield:
Sure, try a few different ideas and see what looks best and makes most sense to you. Post the most compelling example here as a comment. If it's a problem lots of people have, and you come up with an elegant solution that justifies the cost of increasing the `script` API surface, I'll certainly consider it.

jessp01:
Hi @bitfield ,

> Post the most compelling example here as a comment.

I already have:) This is a real feature I am working on that made this functionality necessary. Did you look at the code, pull description and screencast I've referenced?

bitfield:
I think I'm missing something somewhere—the code you linked seems to use `MatchRegexp`, as it currently is, but not `MatchRegexpWithLineNumber`, as you propose. Do you have an example of using `MatchRegexpWithLineNumber`? That's what I'm interested in seeing.

bitfield:
I'll close this for now, since it seems to have stalled—if you'd like to pick it up again at any point in the future, please do.

Issue #226, opened 2025-03-18
Add a filter to handle newline-delimited JSON


tho:
I would like to propose adding a new `JQnl` (or `JQLines`) filter to the script package. The filter would process [newline-delimited JSON](https://jsonlines.org/) data by applying a JQ query to each JSON object in the input. It should work with pretty-printed and compact (`jq -c`; one JSON object per line) JSON input.

Newline-delimited JSON is widely used for streaming data where each line represents a self-contained JSON object. The current `JQ` method, unlike the `jq` command line utility, only processes a single JSON object from the input, making it difficult to process streams of data.

## Use cases

1. Log Analysis

   Application logs are often output as JSON objects, one per line. With `JQnl`, users could extract and transform specific fields from log files, filtering for specific log levels, error messages, etc.

2. Processing Paginated API Results

   When dealing with APIs that return paginated results, `JQnl` would allow processing paginated reponses as part of a script pipeline.

3. Data ETL Workflows

   For Extract-Transform-Load workflows where each record is a separate JSON object, the method would streamline the processing of large data sets by applying transformations to each record in the stream without loading everything into memory at once or rather before applying the `JQ` filter.

## Concrete example for the Log Analisys use case

Extrct all warning and error messages from a JSON log file, e.g. output of `slog`.

`/tmp/log.json` - The mixed compact-prettyprint-compact format is intentional for illustration purposes.
```json
{"time": "2025-03-17T18:04:26.534789-07:00", "level": "INFO", "msg": "info message"}
{
        "time": "2025-03-17T18:04:26.534946-07:00",
        "level": "WARN",
        "msg": "warn message"
}
{"time": "2025-03-17T18:04:26.534953-07:00", "level": "ERROR", "msg": "error message"}
```

`jq` - command line utility for reference
```sh
$ cat /tmp/log.json | jq 'select(.level=="WARN" or .level=="ERROR") | .msg'
"warn message"
"error message"
```

`JQ` - `script.Stdin().JQ(os.Args[1]).Stdout()`
```sh
$ cat /tmp/log.json | ./scriptJQ 'select(.level=="WARN" or .level=="ERROR") | .msg'
$ # no output, since `JQ` only processes the first JSON object
```

`JQnl` - `script.Stdin().JQnl(os.Args[1]).Stdout()`
```sh
$ cat /tmp/log.json | ./scriptJQnl 'select(.level=="WARN" or .level=="ERROR") | .msg'
"warn message"
"error message"
```

[...]

PR #227, opened 2025-03-19
Change JQ to process newline-delimited JSON

tho:
Update `JQ` filter to handle [newline-delimited JSON (NDJSON)](https://jsonlines.org).

Fixes https://github.com/bitfield/script/issues/226.

## Implementation Notes

- Used decoder's `More()` method for cleaner iteration over JSON stream
- Preserved existing code style with short variable names
- Preserved error handling approach with early returns
- Preserves `interface{}` usage instead of `any`
- No return value checks for `fmt.Fprintln`
- Direct return of external package errors without wrapping
- Wrote individual tests instead of table driven tests

## Example

Analyzes logs to count and display frequency of different log levels from newline-delimited JSON input.

```
cat log.json | ./goscript.sh -c 'script.Stdin().JQ(".level").Freq().Stdout()'
3 "INFO"
2 "WARN"
1 "ERROR"
```

### DIFF
--- a/script_test.go
+++ b/script_test.go
@@ -804,12 +804,88 @@ func TestJQHandlesGithubJSONWithRealWorldExampleQuery(t *testing.T) {
 	}
 }
 
+func TestJQWithNewlineDelimitedInputAndFieldQueryProducesSelectedFields(t *testing.T) {
+	t.Parallel()
+	input := `{"timestamp": 1649264191, "iss_position": {"longitude": "52.8439", "latitude": "10.8107"}, "message": "success"}` + "\n"
+	input += input
+	want := `{"latitude":"10.8107","longitude":"52.8439"}` + "\n"
+	want += want
+	got, err := script.Echo(input).JQ(".iss_position").String()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if want != got {
+		t.Error(want, got)
+		t.Error(cmp.Diff(want, got))
+	}
+}
+
+func TestJQWithNewlineDelimitedInputAndArrayInputAndElementQueryProducesSelectedElements(t *testing.T) {
+	t.Parallel()
+	input := `[1, 2, 3]` + "\n" + `[4, 5, 6]`
+	want := "1\n4\n"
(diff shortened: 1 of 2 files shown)


bitfield:
@tho great work! I've made a few minor tweaks for house style and so on—are you okay with these?

script_test.go
```diff
@@ -875,16 +848,15 @@ func TestJQErrorsWithInvalidInput(t *testing.T) {
 	}
 }
 
-func TestJQWithNewlineDelimitedInputErrorsAfterFirstInvalidInput(t *testing.T) {
+func TestJQProducesValidResultsUntilFirstError(t *testing.T) {
 	t.Parallel()
-	input := `[0]` + "\n" + `[1` + "\n" + `[2]` // missing `]` in second line
-	want := "0\n"
+	input := "[1]\ninvalid JSON value"
+	want := "1\n"
```
tho: I suggest adding another valid JSON line after the "invalid JSON value" line and maybe moving the `want != got` check above the `err == nil` check. I understand that this won't be possible, but it's not obvious to me and I had to double check when reading the test. Making the suggested changes would clarify (to me) and verify that processing stops after the first error whether the pipeline returns an error or not.

```diff
--- a/script_test.go
+++ b/script_test.go
@@ -845,15 +845,15 @@ func TestJQErrorsWithInvalidInput(t *testing.T) {

 func TestJQProducesValidResultsUntilFirstError(t *testing.T) {
        t.Parallel()
-       input := "[1]\ninvalid JSON value"
+       input := "[1]\ninvalid JSON value\n[2]"
        want := "1\n"
        got, err := script.Echo(input).JQ(".[0]").String()
-       if err == nil {
-               t.Fatal("want error from invalid JSON input, got nil")
-       }
        if want != got {
                t.Error(cmp.Diff(want, got))
        }
+       if err == nil {
+               t.Fatal("want error from invalid JSON input, got nil")
+       }
 }
```

tho (commented):
lgtm. Thanks for the quick review!

Issue #229, opened 2025-05-06
Replace String within Multiple Files


sam-mininberg:
Hi! I started using script recently, which I've found very useful. One pattern I've found myself using is making in-place replacements across multiple files like this:

`script.ListFiles("*.txt").ExecForEach("sed -i 's/tpo/typo/g' {{.}}").Wait()`

I think it would be nice to have a "ReplaceInEach" function so that I could replace all sed calls with script equivalents. For example:

`script.ListFiles("*.csv").ReplaceInEach("tpo", "typo")`

Here's a simple implementation without any error handling:

```
// ReplaceInEach reads paths from the pipe, one per line, and replaces within
// each file all occurences of the string search with the string replace.
func (p *Pipe) ReplaceInEach(search, replace string) *Pipe {
	return p.FilterScan(func(file string, w io.Writer) {
		File(file).Replace(search, replace).WriteFile(file)
	})
}
```
Please let me know what you think. Thanks!

bitfield:
Thanks @sam-mininberg, great suggestion!

I can see where this would be useful, but I'm not sure how well it fits into the existing model of "everything is a pipe" that `script` uses. There's no _output_ from `ReplaceInEach` that could be used by subsequent pipeline stages, as far as I can work out.

This might be better implemented as a custom `Filter` function, I think.

sam-mininberg:
Thanks for your quick reply, @bitfield! 

You make a good point about the output from `ReplaceInEach`. Perhaps it could return the number of replacements made in each file, in the same format as `Freq`? That way one could verify that the expected replacements were made. A subsequent `Filter` could then be applied, for example, to get the list of files in which at least 1 replacement was made. 

No worries if `ReplaceInEach` doesn't make sense for `script` right now. I like your idea of implementing this with `Filter`! I'll work on that and follow up here.

sam-mininberg:
Here's what I came up with:

```
// ReplaceInEach reads paths from the pipe, one per line, and replaces within
// each file all occurences of the string search with the string replace.
// Returns the paths prefixed with the number of replacements made in each.
func (p *Pipe) ReplaceInEach(search, replace string) *Pipe {
	return p.Filter(func(r io.Reader, w io.Writer) error {
		scanner := newScanner(r)
		for scanner.Scan() {
			file := scanner.Text()
			s, err := File(file).String()
			if err != nil {
				return err
			}
			count := strings.Count(s, search)
[...]

Issue #137, opened 2022-10-27
WriteFile occassionaly truncates files entirely


klauern:
I am trying to write a simple `sed -i` script using this, but it seems that when I run it on the same path that I am reading it from, occasionally the file is completely truncated instead of written with the piped changes.

```go
func Replace(path, from, to string) error {
	pipe := script.File(path)
	file, err := pipe.Replace(from, to).WriteFile(path)
	if err != nil {
		fmt.Println(path, err)
		return fmt.Errorf("error writing file %s: %w", path, err)
	}
    return nil
}
```

I can make this work by pulling the write operation out of the pipe:

```go
func Replace(path, from, to string) error {
	pipe := script.File(path)
	file, err := pipe.Replace(from, to).String()
	if err != nil {
		fmt.Println(path, err)
		return fmt.Errorf("error writing file %s: %w", path, err)
	}
	return os.WriteFile(path, []byte(file), 0644)
}
```

klauern:
What's frustrating is that this isn't the case **all** the time, so it does work for a good percentage, so there is some kind of weird race condition that is causing it not to work.  I am not running any of these calls in parallel, so I am pretty sure it's how the file is handled in `bitfield/script` itself.

bitfield:
I'm not sure it's safe to write to a file at the same time as reading from it—or rather, it's _safe_, but might give you puzzling results, as you've experienced. The reads and writes will be buffered, which means the bits don't necessarily show up on disk when you expect them to.

klauern:
I was thinking that, too.  I thought maybe the `Wait()` command would let me do that, but since it doesn't return a `*Pipe`, that seemed like a dead-end.

andrew-werdna:
I've encountered the same issue as well. Filters run concurrently as stated in the readme, and it doesn't look like there is an option to enforce consecutive execution. I was trying to do the exact same thing as you @klauern and I solved it the same way too.

zhqqqy:
> Another option that avoids reading the whole file into memory would be to pipe it to a temporary file, and then once the pipe has completed, rename the temporary file to the original file.

yes, i used `ioutil.TempFile("", "temp.txt")`  to create a temporary file. Perhaps this could be encapsulated into an internal method to be used with Replace function ? Like `sed -i`.

PR #230, opened 2025-05-16
added README example that makes replacements in multiple files

sam-mininberg:
Addresses #229

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,28 @@ script.Echo("a\nb\nc").FilterScan(func(line string, w io.Writer) {
 // scanned line: "c"
 ```
 
+Alternatively, we can use a `Filter` function that returns a `string`. For example, let's fix typos we've made in the CSV files in our working directory:
+
+```go
+script.ListFiles("*.csv").FilterLine(func(file string) string {
+	search := "typ"
+	replace := "typo"
+	s, err := File(file).String()
+	if err != nil {
+		return fmt.Sprintf("%s %s", file, err)
+	}
+	count := strings.Count(s, search)
+	_, err = Echo(strings.ReplaceAll(s, search, replace)).WriteFile(file)
+	if err != nil {
+		return fmt.Sprintf("%s %s", file, err)
+	}
+	return fmt.Sprintf("%s %d", file, count)
+})
+// Output:
+// a.csv 3
+// b.csv 0
+```
+
 And there's more. Much more. [Read the docs](https://pkg.go.dev/github.com/bitfield/script) for full details, and more examples.
 
 # A realistic use case


bitfield:
Nice work @sam-mininberg!

README.md
```diff
@@ -220,6 +220,28 @@ script.Echo("a\nb\nc").FilterScan(func(line string, w io.Writer) {
 // scanned line: "c"
 ```
 
+Alternatively, we can use a `Filter` function that returns a `string`. For example, let's fix typos we've made in the CSV files in our working directory:
```
bitfield, line 223: I wonder if referencing CSV files is a red herring here. Readers might think this example is something to do with CSV parsing. Should we just talk about making arbitrary edits to all files matching a pattern?

README.md
```diff
@@ -220,6 +220,28 @@ script.Echo("a\nb\nc").FilterScan(func(line string, w io.Writer) {
...
+
+```go
+script.ListFiles("*.csv").FilterLine(func(file string) string {
+	search := "typ"
+	replace := "typo"
```
bitfield, line 228: Maybe we should use completely different search and replace strings, as this particular search and replace would not be idempotent: if you keep running it, "typo" will become "typoo", and so on.

README.md
```diff
@@ -220,6 +220,28 @@ script.Echo("a\nb\nc").FilterScan(func(line string, w io.Writer) {
...
+	s, err := File(file).String()
```
bitfield, line 229: Reading the whole file into a single string is a bit limiting—and unnecessarily so when we only need to read it line by line. What about using `script`'s own `Replace` method, if that's possible here?

README.md
```diff
@@ -220,6 +220,28 @@ script.Echo("a\nb\nc").FilterScan(func(line string, w io.Writer) {
...
+	if err != nil {
+		return fmt.Sprintf("%s %s", file, err)
```
bitfield, line 231: Maybe it's better after all to use `Filter`, then, since that can return an error and stop the pipe if necessary; something doesn't feel right about just flattening it like this and carrying on. What do you think a sensible text-replacement program _should_ do if it hits an error with some file? What would the equivalent `sed` command do?

README.md
```diff
@@ -220,6 +220,28 @@ script.Echo("a\nb\nc").FilterScan(func(line string, w io.Writer) {
...
+	}
+	count := strings.Count(s, search)
```
bitfield, line 233: It's nice to have a count of the number of replacements, but not essential—and the key to a good example is eliminating all non-essential detail. I don't deny that we probably would want a real `script` text-replacement program to do this, but examples like this are more effective when they're as brief and focused as possible, wouldn't you say?

README.md
sam-mininberg: Sorry for disappearing for awhile! Yes I think that's better. I'll make that change.

README.md
sam-mininberg: Good point! I'll change it to "helo" --> "hello".

README.md
sam-mininberg: That's a good point! I originally chose to read in the whole file like this so that I could use `strings.Count`. But I agree with your other comment that the counts are not essential to this example. I'll update my snippet to use `Replace` instead.

README.md
sam-mininberg: I created a test script to test what the equivalent `sed` would do:
```
$ bash test.sh
1.txt before sed: helo world!
3.txt before sed: helo there!
Running "sed -i 's/helo/hello/g' 1.txt 2.txt 3.txt"
sed: can't read 2.txt: No such file or directory
1.txt after sed: hello world!
3.txt after sed: hello there!
```
Even though `sed` doesn't behave this way, I do agree with you that it doesn't feel right to continue after encountering an error. 

It seems like two approaches are:

- Use `Filter` and stop the pipe at the first error
- Use `FilterLine` and put any error messages on the pipe

With all of this in mind, which do you prefer? Or perhaps there is a third approach I'm not considering?

bitfield: Well, you're the customer: as the person who wants to replace text in multiple files, which behaviour would be more convenient or natural to you?

sam-mininberg: Stopping at the first error seems more natural to me. I'll update my implementation accordingly.

README.md
sam-mininberg: Quick update: I'm working on using `Replace` but am running into #137. I've yet to come up with an elegant example that works around the issue, but I'll keep at it.

lazysegtree:
@sam-mininberg are you still working on this?

I have a related issue
- https://github.com/bitfield/script/issues/236

I think this kind of real-world, and slighly-complex examples should be put in the `examples` section.

@bitfield what do you think?

PR #235, opened 2025-10-25
Add timeout support for Exec and ExecForEach commands (fixes #225)

### LINKED ISSUE #225: Allow timeout for Exec to avoid hanging
It would be nice to be able to specify a timeout to avoid commands hanging during execution.

Maybe something like the existing WithEnv(), but instead WithTimeout() or WithWaitDelay() to support the new (Go 1.20) WaitDelay?

https://pkg.go.dev/os/exec#Cmd.WaitDelay

ehsundar:
## Problem

`Exec` and `ExecForEach` commands can run indefinitely, causing resource leaks and hanging CI/CD pipelines with no way to set timeouts.

## Solution

Adds `WithTimeout(duration time.Duration)` method for both `Exec` and `ExecForEach` commands.

```go
// Basic usage
script.NewPipe().WithTimeout(5*time.Second).Exec("sleep 10").Wait()

// ExecForEach with timeout
script.Echo("file1\nfile2").WithTimeout(2*time.Second).ExecForEach("sleep 5").Wait()
```

## Realistic Example: CI/CD Pipeline
```go
// Build with timeout
script.Exec("go build ./...").WithTimeout(2*time.Minute).Wait()

// Test with timeout  
script.Exec("go test ./...").WithTimeout(5*time.Minute).Wait()

// Security scan with timeout
script.Exec("gosec ./...").WithTimeout(1*time.Minute).Wait()
```

## Why This is Better

- **Simple API**: `WithTimeout()` follows existing patterns
- **No external deps**: Built into the library
- **Reliable**: Uses `context.WithTimeout` and `exec.CommandContext`
- **Backward compatible**: Zero timeout means no timeout

Fixes #225

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,14 @@ script.Exec("ping 127.0.0.1").Stdout()
 
 Note that `Exec` runs the command concurrently: it doesn't wait for the command to complete before returning any output. That's good, because this `ping` command will run forever (or until we get bored).
 
+If you need to prevent a command from running indefinitely, you can set a timeout:
+
+```go
+script.NewPipe().WithTimeout(5*time.Second).Exec("potentially-hanging-command").Wait()
+```
+
+If the command doesn't complete within the timeout, it will be terminated and an error will be set.
+
 Instead, when we read from the pipe using `Stdout`, we see each line of output as it's produced:
 
 ```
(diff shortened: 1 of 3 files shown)


bitfield:
Thanks for the PR. What do you think about the comments on #235 225 about letting users inject a context, and also about using `WaitDelay` on the command? I don't think we've yet reached a consensus about how this feature should work, so I think we need to do that before attempting to implement anything.

bitfield:
We now have the context mechanism introduced in #244, so I'll close this for now, but many thanks for the contribution @ehsundar. In future we could think about adding an explicit `WithTimeout` that uses context to achieve it, so this API may end up in the package after all.

PR #240, opened 2026-01-27
add goscript exacutable

prochac:


prochac:
The Go interpreter no-one asked for :P

lazysegtree:
Can you explain what problem this PR is trying to solve, and what is the use case of this?

cmd/goscript/main.go
```diff
@@ -0,0 +1,182 @@
...
+	cmd.Dir = tmpDir
+	if _, err := cmd.Output(); err != nil {
+		return err
+	}
+	cmd = exec.Command(goBinPath, "get", "github.com/bitfield/script")
+	cmd.Dir = tmpDir
+
+	if _, err := cmd.Output(); err != nil {
+		return err
+	}
```
lazysegtree, line 62: What the? Why are we doing all this inside a go code? This is extremely slow, and too much work being done on every single run.

This looks like the worst roundabout way to implement command

If you are doing all that, then users might as well setup a time go project themselves.

lazysegtree (changes requested):
Right, no one asked for this. No one needs this either. And the implementation is unbelievably ugly.

cmd/goscript/main.go
lazysegtree: The rest of the code is so fragile and bad, its hard to look at. I don't even want to spend my time understanding it.

prochac: The whole point of doing it in Go code is that it can be used with `go tool` command. That can't be done with bash script

prochac:
> Can you explain what problem this PR is trying to solve, and what is the use case of this?

The `goscript.sh` from the README.md works only locally. Precisely, only on those machines, that have it in their PATH. When the "interpreter" is written in Go, it can be used with `go tool` command, tracked in `go.mod`, and used by anybody when used in a project.
Also, the installation is much Go-like than wget or copy-paste from a website.

bitfield:
@prochac, thanks for submitting this! It looks really interesting. Here's what I suggest:

1. First, publish this tool as a separate project of its own—it's yours and you should have control of it.
2. If it proves to be helpful and useful, it's certainly something I'd consider adding a link to from the `script` README, so that users can be aware of it.
3. It doesn't need to be part of _this_ project, but if you do want to contribute it, let's talk about that further down the line.

I recommend you take a look at the [CONTRIBUTING](https://github.com/bitfield/script/blob/master/CONTRIBUTING.md) guide in general to help increase the chances of your contributions being accepted—for example, we tend to start with an issue first and then move to a PR only once the shape of a solution is agreed.

While @lazysegtree of course has the right to comment on anything here, as it's a public repo, be aware that he's not a maintainer and doesn't speak for the project. And @lazysegtree, I suggest you review the [Code of Conduct](https://github.com/bitfield/script/blob/master/CODE_OF_CONDUCT.md) to get a sense of how we tend to interact with each other here: I'd sum it up as “friendly, respectful, and collegial”.

lazysegtree:
Sorry, lost my temper there. I will keep in mind to comment objectively and remove my subjective feelings.

PR #241, opened 2026-06-03
bump to go1.26.4 for GO-2026-5037, GO-2026-5039

bitfield:


mhutchinson:
This change forces all downstream projects to bump their version of go to 1.26.4 (especially if they use dependabot), e.g. https://github.com/transparency-dev/tessera/pull/996. I don't believe you're using any go 1.26 features that require this propagation.

Another way to solve this problem of the CVEs in the core libraries is to use the `toolchain` directive instead: https://medium.com/@alif_biswas/why-your-go-mod-has-two-versions-go-and-toolchain-go-996f8065482e

Are you open to reducing the go min version and cutting another release?

bitfield:
Hey, @mhutchinson, thanks for the report!

I did wonder about setting the `toolchain` instead, but I'm not exactly sure what difference it makes. Can you enlighten me? If the project must be built with the standard library from Go 1.26.4, I don't know how you can do that with a Go version less than 1.26.4. Aren't dependents forced to use Go 1.26.4 either way?

mhutchinson:
Hey @bitfield thanks for the response. Always a bit of a long shot commenting on closed PRs!

The idea is that the go version in the go.mod file is the min version of the Go API that you need to compile: https://go.dev/doc/modules/gomod-ref#go-syntax. I don't believe you need any API changes that were introduced in 1.26.4 (there shouldn't have been any!). However, you need the fix for the CVEs. Those were included in 1.26.4, and also ported to 1.25.11.

What I suspect you can get away with to maintain the most flexibility for your users is:

```
go 1.25.0

toolchain go1.25.11
```

That should indicate that your code requires API features in go 1.25, but your compilation and tests should use the non-vulnerable go version.

bitfield:
Sorry, @mhutchinson, but there doesn't seem to be any way to do this—or I haven't found it. With either `go 1.25.11` or `toolchain go1.25.11`, or both, the latest `govulncheck` still reports the two CVEs. Maybe the vulnerability database just hasn't been updated for the backports yet?

mhutchinson:
Hi @bitfield I've created #242 and set the toolchain to 1.26.4. govulncheck passes for me on the repo with a 1.26.0 go toolset in my paths, where the 1.25.11 didn't. Maybe that fixes it?

PR #242, opened 2026-06-10
Lowered go.mod to 1.25.0

mhutchinson:
This restores support for consumers still using Go 1.25. The previous bump to `go 1.26.4` forced all consumers of this library to use Go 1.26 or newer.
The codebase does not use any Go 1.26-specific features, making the `1.26` language requirement unnecessary.

`go1.25.11` contains the same critical security patches as `go1.26.4` (fixing vulnerabilities in `crypto/x509`, `net/textproto`, and `mime`).
Go's toolchain auto-switching (introduced in Go 1.21) will automatically download and use `go1.25.11` for developers working on this module if their local toolchain is older, ensuring they build with the secure version without forcing a major Go upgrade.


mhutchinson:
I also ran go mod tidy, which explains the cleanup in `go.sum`

bitfield:
Nice one @mhutchinson, thanks! I'll use `toolchain` for future security updates now that I know the difference it makes.

PR #243, opened 2026-06-19
docs: add example to count ERROR lines in a log file (#236)

### LINKED ISSUE #236: Add examples/ in the repo
We could add around 5-10 common usage examples in `examples/` directory with this format

file - descriptive_name.go

content - 

```
<go code>
/* 
<Detailed explanation of what program does if needed>

<Equivalent shell script>
*/
```
- Makes it easy for new users to understand the API
- Makes it easy for users to implement a solution for them, by providing them a reference of real world use cases
- It allows providing larger examples that cannot fit in ReadMe 
- It allows providing more examples, as Readme cannot fit too many examples.
- Its standard practive

Bonus
- We could add some sort of check in GitHub Actions to ensure all examples runs. That will be a good intergeration test.

shreyaabaranwal:
Hi! Following up on #236 — here's a first example to get the examples/ directory started.

I went with something straight out of everyday ops life: counting ERROR lines in a log file. Anyone who's stared at logs during an incident knows this one — it's basically `grep ERROR app.log | wc -l`, but here it's done with script:

    script.File(logFile).Match("ERROR").CountLines()

I really like how cleanly the pipe-style API reads here — it maps almost one-to-one to the shell command, which I think makes it a nice, relatable first example for newcomers.

I've included a small sample `examples/app.log` so it runs out of the box, and verified everything locally with `go run`, `gofmt -l`, and `go vet` (all clean).

Happy to add a couple more examples in the same spirit if you'd like — just wanted to start with one and get your feedback on the format first.

Closes #236

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ For something a bit more challenging, let's try counting the number of lines in 
 numErrors, err := script.File("test.txt").Match("Error").CountLines()
 ```
 
+You can find a complete, runnable version of this in [examples/count_errors.go](examples/count_errors.go), along with other examples in the [`examples/`](examples) directory.
+
 But what if, instead of reading a specific file, we want to simply pipe input into this program, and have it output only matching lines (like `grep`)?
 
 ```go
--- a/examples/count_errors.go
+++ b/examples/count_errors.go
@@ -1,0 +1,30 @@
+// This program reads a log file, filters only the lines containing "ERROR", and prints the count of those lines.
+// It uses the github.com/bitfield/script library to handle the file reading, matching, and counting.
+//
+// Equivalent shell command:
+// grep ERROR app.log | wc -l
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/bitfield/script"
+)
+
+func main() {
+	// Check if examples/app.log exists, otherwise fallback to app.log.
+	logFile := "examples/app.log"
+	if _, err := os.Stat(logFile); os.IsNotExist(err) {
+		logFile = "app.log"
+	}
+
+	count, err := script.File(logFile).Match("ERROR").CountLines()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error reading log file: %v\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("Number of ERROR lines: %d\n", count)
+}
--- a/examples/count_errors.go
+++ b/examples/count_errors.go
@@ -1,3 +1,5 @@
+//go:build ignore
+
 // This program reads a log file, filters only the lines containing "ERROR", and prints the count of those lines.
 // It uses the github.com/bitfield/script library to handle the file reading, matching, and counting.
 //
--- a/examples/count_errors.go
+++ b/examples/count_errors.go
@@ -1,32 +1,26 @@
 //go:build ignore
 
-// This program reads a log file, filters only the lines containing "ERROR", and prints the count of those lines.
-// It uses the github.com/bitfield/script library to handle the file reading, matching, and counting.
+// Counts the lines containing "ERROR" in a log file.
+//
+// Run it from the repository root:
+//
+//	go run examples/count_errors.go
 //
 // Equivalent shell command:
-// grep ERROR app.log | wc -l
-
+//
+//	grep ERROR examples/app.log | wc -l
 package main
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/bitfield/script"
 )
 
 func main() {
-	// Check if examples/app.log exists, otherwise fallback to app.log.
-	logFile := "examples/app.log"
-	if _, err := os.Stat(logFile); os.IsNotExist(err) {
-		logFile = "app.log"
-	}
-
-	count, err := script.File(logFile).Match("ERROR").CountLines()
+	count, err := script.File("examples/app.log").Match("ERROR").CountLines()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Error reading log file: %v\n", err)
-		os.Exit(1)
+		panic(err)
 	}
-
 	fmt.Printf("Number of ERROR lines: %d\n", count)
 }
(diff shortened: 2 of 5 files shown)


bitfield:
Great contribution, @shreyaabaranwal, thanks a lot!

It might be a nice idea to tie these examples in more directly with the ones in the README, mightn't it? For example, we start with:

```go
contents, err := script.File("test.txt").String()
```

and work our way through things like counting error lines (to your point). What if we made each of these examples a standalone program, and then we could link from the README to the full listing? If you feel like adding more complicated examples, by all means do, but we should probably start with these.

For examples that need to read a specific file, we should probably come up with a general solution to:

```go
	// Check if examples/app.log exists, otherwise fallback to app.log.
```

My instinct is that people would naturally run `go run examples/count_errors.go`, for instance, which would mean the path we're opening should be `examples/app.log`. We want to minimise any code in the example that's not directly about `script`, purely because that makes it easier for people to see what the example is actually about.

shreyaabaranwal:
Thanks for the feedback! Reworked both examples:

- Dropped the file-existence fallback — they now open examples/app.log (and examples/servers.csv) directly, since people run them from the repo root with `go run examples/count_errors.go`.
- Stripped out everything that wasn't about script itself, so the pipeline line stands on its own.
- Linked them from the README right after the matching "count Error lines" snippet, following the existing progression.

Verified locally with go run, gofmt -l, and go vet (all clean). Happy to add more in the same style once the format looks right.

examples/count_errors.go
```diff
@@ -0,0 +1,26 @@
+//go:build ignore
```
bitfield, line 1: I'm wondering if we actually need this. I mean, it's absolutely correct in the sense that this file isn't part of the `script` package. On the other hand, one thing about example programs is that people will copy and paste them exactly as is—that's what they're for, after all. Including this line would stop their program working, which might be very puzzling for beginners.

Does it do us any harm to omit the build tag in examples?

examples/count_errors.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"fmt"
+
+	"github.com/bitfield/script"
+)
+
+func main() {
+	count, err := script.File("examples/app.log").Match("ERROR").CountLines()
+	if err != nil {
+		panic(err)
```
bitfield, line 23: Again, panicking on errors makes sense for an example because we don't want to clutter up the program with error handling paperwork. But, again, I've learned over the years that whatever you show people is exactly what they'll do, word for word. It's no good saying “Obviously for real programs you wouldn't use `panic` here,” because that's _not_ obvious to beginners, and they are precisely the people we're writing for.

Instead, even when it's not directly relevant to the principle I'm demonstrating, I've got into the habit of writing the code that shows what I think is best practice:

```suggestion
		fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
```

README.md
```diff
@@ -47,6 +47,8 @@ For something a bit more challenging, let's try counting the number of lines in
 numErrors, err := script.File("test.txt").Match("Error").CountLines()
 ```
 
+You can find a complete, runnable version of this in [examples/count_errors.go](examples/count_errors.go), along with other examples in the [`examples/`](examples) directory.
```
bitfield, line 50: In my books I use a listing link convention something like this:


```suggestion
(Listing [`count_errors`](examples/count_errors.go))
```

We can use that style for all the other linked examples to come, and perhaps this first time mention the `examples` directory:


```suggestion
You can find a complete, runnable version of this and other examples in the [`examples/`](examples) directory.
```

What do you think?

bitfield (commented):
Amazing work @shreyaabaranwal, thanks! This is super helpful.

I've suggested a couple of presentational tweaks, but otherwise I think this is ready to merge as is. Nice job.

README.md
shreyaabaranwal: I like the (Listing `name`) convention — I'll use the directory-mention version for this first link and that style for the rest, once we settle the layout.

examples/count_errors.go
shreyaabaranwal: Will do — switching to fmt.Fprintln(os.Stderr, err) + os.Exit(1) when I apply the restructure.

examples/count_errors.go
shreyaabaranwal: Good question — I tried it, and removing the build tag does break things: with both count_errors.go and filter_csv.go declaring func main() in the same examples/ directory, `go build ./...` and `go vet ./...` fail with "main redeclared". `go run file.go` handles one file fine, but a directory-wide build compiles them as one package, so the tag was working around exactly that.

To keep them copy-paste-able without the tag, my instinct is to give each example its own subdirectory (examples/count_errors/main.go, examples/filter_csv/main.go) — scales cleanly and `go build ./...` stays happy. But it's your call on the layout. Which would you prefer?

bitfield: Yes, we're working against the grain of Go a little bit here, aren't we? It's actually quite awkward to include a copypastable program in a Go module that doesn't cause a `main` conflict.

Maybe what we should do instead is use Go's [built-in example](https://bitfieldconsulting.com/posts/examples) mechanism. For example (sorry), we could add this to `script_test.go`:

```go
func Example_count_lines() {
	count, err := script.File("examples/app.log").Match("ERROR").CountLines()
	if err != nil {
		panic(err)
	}
	fmt.Printf("Number of ERROR lines: %d\n", count)
}

func Example_filter_csv() {
	// Column splits on whitespace, so turn the comma into a space first,
	// then take the first field (the server name).
	_, err := script.File("examples/servers.csv").Match("DOWN").Replace(",", " ").Column(1).Stdout()
	if err != nil {
		panic(err)
	}
}
```

As you probably know, because the function names start with `Example`, they'll be built as part of the autogenerated documentation, and because they don't include the names of any other identifier, they'll be treated as package-level examples:

Screenshot 2026-06-29 at 13 25 11

We can put the data files in the `testdata` folder—I don't think the public `pkgsite` instance will let you read these, but that's okay. People can copy and paste the code and it'll work.

bitfield:
Hey @shreyaabaranwal, just checking in to see how you're getting on—are you happy with the general direction here? It's ended up being a bit different from your initial vision, but in my experience that always happens—and it's a sign of a healthy design process that it _can_ change direction.

Issue #239, opened 2026-01-15
WithEnv + Exec(ForEach) don't work togheter


prochac:
The `script.Exec` contains this code
```go
args, err := shell.Fields(cmdLine, nil)
if err != nil {
	return err
}
```
that already evaluates the env variables, but `nil` is passed in. So the variables are evaluated as empty string.

https://go.dev/play/p/P73IM0NLNBv

Or am I using it wrong?

prochac:
Or it shouldn't be evaluated at all:
https://go.dev/play/p/XWxxo4eEla_t

lazysegtree:
This is a valid issue. We don't respect the pipe's `env` while parsing shell command passed to `Exec`

```go
fmt.Printf("Started\n")
	env := []string{`MY_ENV=foo`}
	out1, _ := script.NewPipe().WithEnv(env).Exec(`echo $MY_ENV`).String()

	cmd := exec.Command("sh", "-c", `echo $MY_ENV`)
	cmd.Env = env
	out2, _ := cmd.Output()

	out3, _ := exec.Command("sh", "-c", `MY_ENV=foo; echo $MY_ENV`).Output()
	fmt.Printf("out1 = '%s'\n", out1)
	fmt.Printf("out2 = '%s'\n", out2)
	fmt.Printf("out3 = '%s'\n", out3)

	os.Setenv("MY_ENV", "foo")
	out4, _ := script.NewPipe().WithEnv(env).Exec(`echo $MY_ENV`).String()
	fmt.Printf("(After setting env) out4 = '%s'\n", out4)

```

This prints

```bash
➜  examples git:(temp) ✗ go run env.go
Started
out1 = '
'
out2 = 'foo
'
out3 = 'foo
'
(After setting env) out4 = 'foo
'
➜  examples git:(temp) ✗```
```

lazysegtree:
The way to fix it is passing a function to get env in `shell.Fields` as you pointed out. 

<details><summary>Proof of concept</summary>
<p>

```diff
diff --git a/script.go b/script.go
index d7d1bc3..2efebac 100644
--- a/script.go
+++ b/script.go
@@ -428,7 +428,7 @@ func (p *Pipe) Error() error {
 // instead be redirected to a supplied writer, using [Pipe.WithStderr].
 func (p *Pipe) Exec(cmdLine string) *Pipe {
 	return p.Filter(func(r io.Reader, w io.Writer) error {
-		args, err := shell.Fields(cmdLine, nil)
[...]

PR #245, opened 2026-07-31
Add Shell method for running commands via the system shell

Dhanalakshmi-D04:
Adds a new `Shell` method (and package-level `Shell` function) that runs
a command line through the operating system's native shell, rather than
manually parsing arguments the way `Exec` does. This fixes the
`WithEnv`/variable-expansion bug described in #239, while leaving `Exec`
itself unchanged for backward compatibility, per the discussion on the
issue.

--Implementation

- Unix-like systems: the command line is passed to `sh -c` unmodified,
  so `$VAR`-style expansion, quoting, and other shell syntax are handled
  by the real shell.
- Windows: the command line is passed to `cmd /C`, following the
  direction suggested in the issue ("give the command line to CMD.exe
  so that it's processed in the same way that it would be if the user
  had typed it into a terminal"). Note that Windows uses `%VAR%` syntax
  rather than `$VAR`, since that's native to `cmd.exe`.
- The platform-specific command construction lives in `shell_unix.go`
  and `shell_windows.go`, using the same `//go:build` pattern already
  used for `script_unix_test.go` / `script_windows_test.go`.

--Testing
- Unix implementation tested on Fedora Linux — all tests pass, including
  a test confirming `WithEnv` variables are now correctly expanded
  (`TestShell_ExpandsEnvironmentVariablesSetViaWithEnv`), directly
  verifying the fix for the original bug.
- Windows implementation tested on Windows 11 (PowerShell) — all tests
  pass, including the Windows equivalent using `%VAR%` syntax.
- Full existing test suite (`go test ./...`) passes on both platforms
  with no regressions.

--Docs

- Added doc comments for `Shell` (package function and method).
- Updated the README's Sources and Filters tables.

I couldn't find a `doMethodsOnPipe` stress-test helper mentioned in
CONTRIBUTING.md anywhere in the current codebase — let me know if it's
been renamed or removed, and I'll update accordingly.

Happy to adjust naming, structure, or behavior based on feedback.


bitfield:
> I couldn't find a doMethodsOnPipe stress-test helper mentioned in
CONTRIBUTING.md

Good spotting—that's out of date. I'll update the contributing guide.

bitfield:
Is it possible to implement this in terms of `Exec`? That would mean we don't have to repeat all the command-setup code and keep these two methods in sync. For example, I recently merged #244, so `Exec` now uses a context if there is one on the pipe. If you used `Exec` to write `Shell`, you wouldn't have to do anything special to pick up that change; you'd get it for free.

Dhanalakshmi-D04:
Good point on the duplication — that makes sense, especially with the context support in #244 that Shell would otherwise miss out on. I'll refactor so Shell builds on Exec's command-setup logic instead of duplicating it, and update the PR shortly.

Dhanalakshmi-D04:
Done — extracted the shared command-running logic (stdin/stdout/stderr, WithEnv, start/wait, error handling) into an unexported runCmd helper used by both Exec and Shell. Shell now also uses exec.CommandContext, so it picks up context support from #244 automatically. Verified with go build, go vet, and the full test suite on both Linux and Windows — all clean. Pushed as a separate commit so the diff is easy to follow alongside the original.

bitfield:
@Dhanalakshmi-D04 nice work—I wonder if we could also actually implement `Shell` by *calling* `Exec`—that way we wouldn't need a third function. What do you think?

It also struck me as a shame that we have to create two new source files just to get the shell invocation command line. Would you be in favour of switching on `runtime.GOOS` instead for the value of this one variable?

Dhanalakshmi-D04:
Good ideas, thanks — I'll implement both:

- Shell calling Exec directly (building the shell-invocation command line and letting Exec's existing parsing/execution handle the rest), removing the separate runCmd helper.
- Switching on runtime.GOOS for the shell/flag choice instead of the two build-tagged files.

I'll need to double check the quoting so the original command line survives intact as a single argument to sh -c / cmd /C through Exec's existing argument parsing — will test thoroughly on both platforms before pushing.

Dhanalakshmi-D04:
Hi @bitfield ,

Following up on this PR — Shell now builds the shell-invocation command line and delegates to Exec, so we no longer need separate shell_unix.go/shell_windows.go files.

Tests pass on Linux and Windows. On Windows I do occasionally see an unrelated testscript cleanup error after tests pass (unlinkat ...\args.exe: Access is denied) — I checked and it also happens on main before this change, so it looks like a pre-existing Windows flake in testscript's temp-dir cleanup, not related to this PR.

Ready for review whenever you get a chance!

bitfield:
I see now why you used a third function `runCmd`—because `Exec` splits its command line using `shell.Fields`, which is the problem we were originally trying to address!

I think your approach was quite right: re-use as much common code as you can between `Exec` and `Shell`, but one thing they mustn't share is calling `shell.Fields`. Of course we also don't want to escape any quotes and so on—the point of `Shell` is to interfere as little as possible with the command line the user supplies.

Sorry for misdirecting you—I just didn't spend enough time and energy on thinking really hard about the problem and what we were really trying to do.

Maybe this is a good opportunity to have a re-think about how we handle command execution in general. There's probably a case for three different methods:

* `ExecCommand` — takes each arg separately, like `exec.Command`
* `Exec` — splits its input command line with `shell.Fields` (and can pass the result to `ExecArgs`) — we'll deprecate this
* `Shell` — passes its command line unmodified to `ExecCommand` with the appropriate `shell`, `flag` args in front.

What do you think of this scheme?

Dhanalakshmi-D04:
That makes sense — I see the issue now: even with careful quoting,routing through Exec meant cmdLine was still passing throughshell.Fields, which is exactly the re-interpretation Shell is meant toavoid.

The three-method split makes sense to me:

- ExecCommand(name, args...) — no parsing, just runs the command directly; this is where the shared stdin/stdout/stderr/env/context/error-handling logic lives.
- Exec(cmdLine) — keeps splitting with shell.Fields as before, then calls ExecCommand with the result. Happy to mark it Deprecated per your suggestion.
- Shell(cmdLine) — calls ExecCommand("sh", "-c", cmdLine) directly (or cmd /C on Windows), so cmdLine passes through completely untouched,no quoting needed.

I'll implement this and test on both platforms.

Dhanalakshmi-D04:
I've gone ahead and implemented the three-method split:

ExecCommand(name, args...) — new method, no parsing at all; holds the shared stdin/stdout/stderr/env/context/error-handling logic.
Exec(cmdLine) — now a thin wrapper that splits with shell.Fields and delegates to ExecCommand; marked Deprecated in favour of ExecCommand and Shell.
Shell(cmdLine) — now calls ExecCommand(shell, flag, cmdLine) directly, so cmdLine passes through completely unmodified — no parsing, no quoting needed anywhere.

One small behavioural note I wanted to flag rather than leave unsaid: previously, a shell.Fields parse error inside Exec only surfaced once the pipe was read. Now Exec sets the error immediately via p.WithError(err) before returning. Existing tests still pass, but please let me know if this is a concern.

I've tested this with go build, go vet, gofmt -l, and the full test suite on both Linux and Windows, and also manually checked that Shell still handles a command line with an embedded single quote correctly. Everything looks clean on my end, but please do take a close look — I'd really appreciate you pointing out anything I've missed or gotten wrong, or any changes you'd like to see.

bitfield:
Great! Actually, let's leave `Exec` undeprecated for now—I think we'd be wise to let the new APIs settle for a while before nudging people to adopt them.

> previously, a shell.Fields parse error inside Exec only surfaced once the pipe was read.

That's a good point: is it possible to add a test that detects this problem? That way, you can prove that your fix works! (I'm sure it does, but it'll be a useful regression check.)

Dhanalakshmi-D04:
Done on both counts:

- Removed the `Deprecated` marker from `Exec` — leaving it as-is for now while the new APIs settle in.
- Added `TestExec_SetsErrorImmediatelyOnInvalidCommandLineSyntax`, which calls `Exec` with a syntactically invalid command line and checks `p.Error()` immediately, with no read of the pipe in between. It passes, confirming the fix works as intended.
Verified clean on both Linux and Windows (`go build`, `go vet`,`gofmt -l`, full test suite).

bitfield:
Thanks a lot, @Dhanalakshmi-D04, great work!

